PR #31, opened 2026-06-02
Feat/poc

### LINKED ISSUE #4: Core Orchestrator and Context Engine
# Core Orchestrator and Context Engine
## 📋 Overview

**User Story:** 
As **Platform Peter** (Platform Engineer), I want a foundational core service that acts as a plug-and-play context engine, so that I can easily connect external AI tools, aggregate their data into a unified catalog (linking models, datasets, endpoints, and users), and expose this via OpenMFP without tightly coupling the core platform to specific vendor implementations.

As **AI Engineer rAIner**, I want to open the Naira OpenMFP dashboard and see a unified, read-only view of all discovered AI assets and their relationships, so I can understand what is available and how it is connected.

---

## 🎯 Context
Currently, AI engineering tools are highly fragmented. Naira aims to resolve this by providing an Internal Developer Platform (IDP) that unifies these tools. Before we can build deep integrations (like the AI Gateway or specific templates), we need a robust, extensible foundation. 

This feature focuses on the absolute core of the Naira backend: a logical engine capable of importing tools and assets, abstracting them, and setting them into context with one another. Because the AI tooling space moves exceptionally fast, this engine must be highly adaptable to new types of assets and relationships without requiring continuous core platform code changes.

---

## 🏗️ Technical Architecture (Logical View)

To keep the core platform decoupled from the underlying storage mechanism (which is evaluated separately), the architecture is defined by its logical capabilities:

### 1. Ingestion Interface/Component
A standardized mechanism (API or event sink) that allows external plugins and connectors to push/pull data into Naira. The ingestion layer must support:
- **Entities:** The core AI assets (e.g., Models, Datasets, Applications).
- **Metadata Aspects:** Attaching additional context to entities (e.g., ownership, costs, eval scores, lineage) without changing the entity's base definition.
- **Relationships:** Defining directed links between entities (e.g., `Application X` -> `uses` -> `Model Y` -> `trainedOn` -> `Dataset Z`).

### 2. Context & Relationship Engine
The logical brain of the platform that stores and resolves the ingested data. 
- It must understand first-class relationships, ensuring that when an AI Engineer looks at an Inference Endpoint, they can easily trace back to the exact Model and Dataset it relies on.
- It must act as a highly extensible catalog where new domain concepts (like a `GuardrailPolicy` or an `Agent`) can be introduced by plugins over time.

### 3. Unified Presentation API
A read-only API layer that queries the Context Engine and serves the mapped entity graph to the OpenMFP frontend. It translates complex backend relationships into a unified structure that the UI can easily render.

---

## 🛠 Functional Requirements

- **Asset Registration:** The system must allow external sources to register AI entities with stable, unique identifiers (e.g., URNs).
- **Metadata Extension:** The system must allow attaching and updating arbitrary metadata (aspects) to existing entities.
- **Relationship Mapping:** The system must support creating, updating, and deleting directional relationships between entities.
- **Query & Discovery:** The system must provide an API to query entities and traverse their relationships (e.g., "Find all Inference Endpoints connected to Model X").
- **UI Read-Only Dashboard:** An OpenMFP micro-frontend that consumes the presentation API to display an asset catalog and visualize the relationships.

---

## 🛡️ Non-Functional Requirements (NFRs)

- **Extensibility:** Introducing a new type of AI asset or a new relationship type must not require recompiling or redeploying the core engine service.
- **Scalability:** The system must efficiently store and resolve relationships for thousands of models, datasets, and endpoints.
- **Tenant Isolation:** Asset visibility and relationship querying must strictly adhere to workspace and tenant isolation boundaries.
- **Resilience:** The failure of an external tool connector must not impact the availability of the core context engine.

---

## ✅ Acceptance Criteria

- [x] A logical API or interface is defined for registering Entities, Aspects, and Relationships.
- [ ] A mock plugin successfully registers a dummy `Model` entity, a dummy `InferenceEndpoint` entity, and a `uses` relationship between them.
- [x] The core engine correctly resolves and returns this relationship when queried.
- [x] The OpenMFP frontend displays the mock assets and explicitly shows the relationship link.
- [ ] An Architecture Decision Record (ADR) is finalized, explicitly defining the technical storage and implementation mechanism (e.g., Entity Catalog via CRDs vs. Extensible Graph DB) that fulfills this logical feature.

libera13:
## Description

PR introduces a custom catalog capable of storing entities and relationships. Includes two example plugins: litellm and mlflow.
Additionally:
- poc-ui
- dev k8s environment to work with plugins 

Reference docs:
- [ADR - 002-build-a-custom-catalog-core](https://github.com/naira-project/architecture/blob/main/adr/002-build-a-custom-catalog-core.md)
- [RFC- Core API design](https://github.com/naira-project/architecture/pull/7)

## Type of Change
- [X] ✨ New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
- [x] `go test ./...` passes

## Checklist

## AI Usage
gpt-5.4 for coding

## Screenshots
image


catalog/internal/httpapi/router.go
```diff
@@ -0,0 +1,179 @@
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	_, _ = w.Write(buf.Bytes())
+}
+
+func writeError(w http.ResponseWriter, err error) {
+	status := http.StatusBadGateway
```
mkorbi: Just to understand this defaults to 502 Bad Gateway for unmatched errors, right? Wouldn't it be conventionally more correct to default to 500; 502 only fits when the failure is genuinely upstream (e.g. a plugin HTTP call)

catalog/internal/plugins/litellm/plugin.go
```diff
@@ -0,0 +1,256 @@
...
+		nodes = append(nodes, appNode)
+
+		if strings.TrimSpace(app.LiteLLMVirtualKey) == "" {
+			continue
+		}
+
+		allowedModels, err := p.fetchAllowedModels(ctx, app.LiteLLMVirtualKey)
+		if err != nil {
+			return pluginapi.IngestionRequest{}, fmt.Errorf("fetching allowed LiteLLM models for app %q: %w", app.Name, err)
+		}
```
mkorbi, line 109: so in that case, if a single key lookup fails it aborts the entire ingestion, doesn't?

ui-poc/README.md
```diff
@@ -0,0 +1,46 @@
+# Getting Started with Create React App
```
mkorbi, line 1: So it doesn't use/integrate right now OpenMFP?

mkorbi (commented):
Overall, lgtm, major caveat: it is just standalone at the moment and not using/integrated in OpenMFP right?

mkorbi:
As discussed today in review, fine for me to have it stand alone for now. Future task: integrate and utilize OpenMFP

catalog/internal/catalog/memory_store.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+type Store interface {
+	ListNodes() []Node
+	GetNode(id NodeID) (Node, error)
+	ListRelations() []Relation
+	UpsertGraph(nodes []NodeClaim, relations []RelationClaim) (int, int, error)
+}
+
+type MemoryStore struct {
```
jagdeepm, line 22: If I understand it , this memory store is where you store all the graph relation and provide mechanism for querying ? In future we could potentially use a Graph DB as a backend so you can rely on existing technology , so does this code provide enough abstraction to make a switch to another Memory Store/ DB to store the graph and relationships ?

catalog/internal/httpapi/router.go
libera13: Yes that made more sense when the idea was for the catalog to act as a proxy. Thanks, I will change that to return 500 by default

catalog/internal/plugins/litellm/plugin.go
libera13: Yes. I thought about being strict here, to get an error if there were some problems in setup. This helps when we create environment were Naira operates. In other scenarios... more production ready, it's better to just return whatever could be fetched

catalog/internal/catalog/memory_store.go
libera13: That's correct the plan is to use a database later for persistence. Thanks to interfaces and the service layer, we can pass different store implementations to the service. So yes, for example, integration with a Postgres store would theoretically only require coding the Postgres store without touching the API or service layer

ui-poc/src/pages/DatasetRegistries.tsx
```diff
@@ -0,0 +1,9 @@
+export default function DatasetRegistries() {
+  return (
+    <div className="flex h-screen items-center justify-center bg-background dark:bg-background-dark-default">
+      <p className="text-sm text-foreground-secondary dark:text-foreground-dark-secondary">
+        Dataset Registries — coming soon.
+      </p>
+    </div>
+  );
```
kocakkaan: I am not sure if we should push this component as a placeholder to Naira, what do you think? Maybe we can delete it altogether and push them when UI for dataset registries is ready.

ui-poc/src/pages/Files.tsx
```diff
@@ -0,0 +1,7 @@
+export default function Files () {
+    return (
+        <div>
+            Placeholder div for showing the data sets and files that this model is attached to 
+        </div>
+    )
+}
```
kocakkaan: Same here, maybe we can delete this one as well and provide it in the next PRs.

ui-poc/src/App.tsx
```diff
@@ -0,0 +1,24 @@
...
+import ModelSpec from './pages/ModelSpec';
+
+function App() {
+  return (
+    <ThemeProvider>
+      <Router>
+        <Routes>
+          <Route path="/" element={<Dashboard />} />
+          <Route path="/model-registry" element={<ModelRegistries />} />
+          <Route path="/dataset-registry" element={<DatasetRegistries />} />
```
kocakkaan: If we would be alright with deleting dataset registries, then we should delete it here too.

catalog/internal/plugins/litellm/plugin.go
mkorbi: ok, got it. Maybe leave here a TODO, so can catch up at a later point

libera13: in [rfc 004 api design](https://github.com/naira-project/architecture/blob/main/rfc/rfc-004-core-api-v1-design.md) is sentence "plugin writes a full snapshot always". 

Let's stick to that. If a plugin returns an error, abort the entire ingestion. I think it's better to see the full data or an error than to see partial data and wonder why something is missing

ui-poc/src/pages/DatasetRegistries.tsx
libera13: done

ui-poc/src/pages/Files.tsx
libera13: done

ui-poc/src/App.tsx
libera13: done

akavel-reply (approved):

PR #44, opened 2026-06-05
Feat/remove stale data

libera13:
## Description

This PR introduces atomic snapshots to the catalog to resolve stale data issues. Any data previously associated with a plugin that is not present in the latest snapshot is automatically pruned.

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [x] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

## AI Usage
gpt-5.4 for coding


catalog/internal/catalog/memory_store.go
```diff
@@ -99,11 +105,15 @@ func (s *MemoryStore) UpsertGraph(nodes []NodeClaim, relations []RelationClaim)
 		existing, ok := s.nodes[id]
 		if ok {
 			existing.ID = id
+			existing.Plugin = pluginName
+			existing.SnapshotID = snapshotID
 			existing.Properties = node.Properties
 			s.nodes[id] = existing
 		} else {
 			s.nodes[id] = Node{
```
akavel-reply: Hm; I was assuming if multiple plugins added the same Node or Relation, this would be recorded, and it would be retained if one of the plugins stops claiming it, but others still do.

I think it would be a good case to add tests that would ensure this.

Let's have a call to talk about this, I'm not currently sure it makes sense to add this PR without that feature.

akavel-reply (changes requested):

docs/plugin-authoring.md
```diff
@@ -0,0 +1,41 @@
...
+- `kind` must not contain `/`
+- `path` may contain `/`
+
+Why: node reads use the route `/v1/nodes/{kind}/*`, so `kind` is one path segment and `path` is the remaining opaque path.
+
+## Current conventions
+
+- Reuse known kinds from `pluginapi/schema.go` when they fit.
+- Keep `kind` stable. Changing it changes node identity.
+- Use `path` for source-specific hierarchy such as `mlflow/model-a` or `litellm/team/app`.
```
akavel-reply: I'm not convinced about recommending a plugin name as prefix, as in `litellm/` or `mlflow/`

catalog/internal/catalog/memory_store.go
libera13: Good idea, I've started the changes to store props in buckets

docs/plugin-authoring.md
libera13: Could you give some examples of what will fit more? You're working on a plugin, so maybe you have something specific in mind?

akavel-reply: It's more that I don't want to anchor our thinking onto this idea yet. I'm fine for now if specific plugins do it like so, but maybe others won't, and I believe it's too early to already codify it as our policy. I agree with "hierarchy" and with `/` for separator, but I don't agree with "source-specific" word here nor with requesting plugin name as prefix. Maybe it'll show up in the future we do want to do this, but I'm not convinced yet and don't want to limit so early the space of possibilities we're exploring.

In other words: I'm ok that in your plugins code you experiment with doing it, I'm not ok with adding it as an explicit policy in docs at this point in time.

Also, it seems to me such a fixed prefix might be redundant with adding a new "kind".

libera13: That makes sense. I've made the change

catalog/internal/httpapi/resources.go
```diff
@@ -3,19 +3,25 @@ package httpapi
...
 type Node struct {
-	Name string `json:"name"`
-	Kind string `json:"kind"`
-	Path string `json:"path"`
-	//FIXME: change to map[string]map[string]string
-	Props map[string]string `json:"props,omitempty"`
+	Name  string               `json:"name"`
+	Kind  string               `json:"kind"`
+	Path  string               `json:"path"`
+	Props []PluginContribution `json:"props"`
```
akavel-reply: would you be ok to rename this type `PluginProps`? this would sound more obvious to me, and easier to remember (more internally consistent/unified naming)

Or, maybe it could not be bad idea to name this field `Aspects`? then we could have:
```go
type Node struct {
    ...
    Aspects []Aspect `json:"aspects"`
}

type Aspect struct {
    Plugin string
    Props map[string]string
}
```

Yet another option I'm pondering is `PluginClaims` - then:
```go
type Node struct {
    ...
    PluginClaims []PluginClaim `json:"pluginClaims"
}

type PluginClaim struct {
    Plugin string
    Props map[string]string
}
```
One thing I like about `PluginClaims` (apart from consistency with the naming from IngestionRequest) is that then a **response with empty `props` would still give useful information in a nicely meaningful way** - specifically, which plugins made the claim that this node/relation exists:
```json
{
    "kind": "app"
    "path": "mycluster/myapp",
    "pluginClaims": [
        {"plugin": "github_scanner"},
        {"plugin": "k8s_scanner"}
    ]
}
```

catalog/internal/httpapi/resources.go
```diff
@@ -139,6 +144,32 @@ func relationName(kind string, from pluginapi.NodeID, to pluginapi.NodeID) strin
 	return fmt.Sprintf("relations/%s/%s|%s", kind, url.PathEscape(nodeName(from)), url.PathEscape(nodeName(to)))
 }
 
+func toSortedSlice(contributions map[string]catalog.PluginContribution) []PluginContribution {
+	if len(contributions) == 0 {
+		return []PluginContribution{}
+	}
```
akavel-reply: I don't think this `if` is needed, the code should work fine without it, I believe?

catalog/internal/httpapi/resources.go
```diff
@@ -139,6 +144,32 @@ func relationName(kind string, from pluginapi.NodeID, to pluginapi.NodeID) strin
...
+
+	// Collect plugin names and sort them for a stable, deterministic output order.
+	pluginNames := make([]string, 0, len(contributions))
+	for pluginName := range contributions {
+		pluginNames = append(pluginNames, pluginName)
+	}
```
akavel-reply: (optional) I think we could drop above loop, and instead first create `result` slice from the `contributions` map in random order, _then_ sort it by the `.Plugin` field. I feel this would make the code somewhat shorter, and remove the need for the variable `pluginNames`.

catalog/internal/catalog/memory_store.go
```diff
@@ -76,71 +89,124 @@ func (s *MemoryStore) ListRelations() []Relation {
...
-				ID:         id,
-				Properties: node.Properties,
+		if !ok {
+			existing = Node{
+				ID:            id,
+				Contributions: make(map[string]PluginContribution),
 			}
+		} else if existing.Contributions == nil {
+			existing.Contributions = make(map[string]PluginContribution)
+		}
```
akavel-reply, line 123: I think we could avoid duplicated `make` and the need for `ok` in above block with:
```go
existing := s.nodes[id]
if existing == nil {
    existing = Node{ID: id}
}
if existing.Contributions == nil {
    existing.Contributions = make(...)
}
```

catalog/internal/catalog/memory_store.go
```diff
@@ -76,71 +89,124 @@ func (s *MemoryStore) ListRelations() []Relation {
...
+		existing, ok := s.relations[key]
+		if !ok {
+			existing = Relation{
+				Kind:          key.Kind,
+				From:          key.From,
+				To:            key.To,
+				Contributions: make(map[string]PluginContribution),
 			}
+		} else if existing.Contributions == nil {
+			existing.Contributions = make(map[string]PluginContribution)
```
akavel-reply: similar comment as above

catalog/internal/catalog/service_test.go
```diff
@@ -47,16 +55,18 @@ func TestListNodesProjectsStoredNode(t *testing.T) {
...
 	assert.Equal(t, "model", node.ID.Kind)
 	assert.Equal(t, "mlflow/fraud-detector", node.ID.Path)
+	contribution, ok := node.Contributions["test-plugin"]
+	require.True(t, ok)
 	assert.Equal(t, map[string]string{
 		"source":      "mlflow",
 		"description": "registry model",
 		"owner":       "risk-platform",
-	}, node.Properties)
+	}, contribution.Properties)
```
akavel-reply: Here and in other tests, maybe we could make a comparison of the whole object with `assert.Equal()`, rather than manually checking field-by-field? I believe `assert.Equal()` is quite powerful for complex types, and can show nice diffs when inequality is detected, and also I'd find such code easier to read. For example:
```console
$ cat main_test.go
package main

import (
        "testing"

        "github.com/stretchr/testify/assert"
)

func TestFoo(t *testing.T) {
        type loose map[string]any
        have := loose{"a": 1, "b": loose{"c": 3, "e": 5}}
        want := loose{"a": 1, "b": loose{"c": 4, "d": 5}}
        assert.Equal(t, want, have)
}

$ go test -v
=== RUN   TestFoo
    main_test.go:13:
                Error Trace:    /Users/m.czaplinski/work/test-assert/main_test.go:13
                Error:          Not equal:
                                expected: main.loose{"a":1, "b":main.loose{"c":4, "d":5}}
                                actual  : main.loose{"a":1, "b":main.loose{"c":3, "e":5}}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -3,4 +3,4 @@
                                  (string) (len=1) "b": (main.loose) (len=2) {
                                -  (string) (len=1) "c": (int) 4,
                                -  (string) (len=1) "d": (int) 5
                                +  (string) (len=1) "c": (int) 3,
                                +  (string) (len=1) "e": (int) 5
                                  }
                Test:           TestFoo
--- FAIL: TestFoo (0.00s)
FAIL
exit status 1
FAIL    test-assert     0.470s
```

catalog/internal/catalog/service_test.go
```diff
@@ -168,3 +182,211 @@ func TestRunAllPluginsUpsertsCollectedGraph(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// 4. Relation survives because litellm still claims it
+	relations = store.ListRelations()
+	require.Len(t, relations, 1)
+	rel = relations[0]
+	require.Len(t, rel.Contributions, 1)
+	assert.Equal(t, map[string]string{"virtual_key": "vk-001"}, rel.Contributions["litellm"].Properties)
+	_, ok := rel.Contributions["mlflow"]
+	assert.False(t, ok)
```
akavel-reply: As I ask in other test too: could you please rewrite those comparison blocks to a single `assert.Equal` comparing with a full expected output? that would make things much easier for me to read and grasp, and compare with previous steps - e.g. here something like:
```go
assert.Equal(t, []Relation{
    {
        Kind: "uses_model", // unchanged
        ...
    },
}, store.ListRelations())
```
(the `// unchanged` comment could help mark those fields which are not expected to have changed)

catalog/internal/catalog/memory_store.go
libera13: `existing` is value type `Node`, do you propose to change `MemoryStore` to keep pointers for nodes?
current
```
type MemoryStore struct {
	mu        sync.RWMutex
	nodes     map[NodeID]Node
	relations map[RelationID]Relation
}
```
proposed?
```
type MemoryStore struct {
	mu        sync.RWMutex
	nodes     map[NodeID]*Node
	relations map[RelationID]*Relation
}
```

Otherwise `if existing == nil {` will not work

catalog/internal/httpapi/resources.go
libera13: Thanks! Changed

catalog/internal/catalog/memory_store.go
akavel-reply: Ah, ok, I didn't notice, you're right. I'm used to working with pointers, so I just assumed the elements are pointers without checking. Given this, I withdraw my comment about `nil` checking - sorry for the noise!

catalog/internal/httpapi/router_test.go
```diff
@@ -137,12 +137,12 @@ func TestRouterServesCurrentEndpoints(t *testing.T) {
 				require.Len(t, payload.Relations, 1)
```
akavel-reply: I was hoping you'd also change the tests in this or any other files, to also just use `assert.Equal()` instead of field-by-field comparison... is there a particular reason you decided to keep the checks unchanged in this file?

catalog/internal/httpapi/resources.go
```diff
@@ -3,19 +3,26 @@ package httpapi
...
 type Node struct {
-	Name string `json:"name"`
-	Kind string `json:"kind"`
-	Path string `json:"path"`
-	//FIXME: change to map[string]map[string]string
-	Props map[string]string `json:"props,omitempty"`
+	Name         string        `json:"name"`
+	Kind         string        `json:"kind"`
+	Path         string        `json:"path"`
+	PluginClaims []PluginClaim `json:"pluginClaimss"`
```
akavel-reply: typo

```suggestion
	PluginClaims []PluginClaim `json:"pluginClaims"`
```

docs/plugin-authoring.md
```diff
@@ -0,0 +1,61 @@
...
+Each execution of a plugin is treated as a single, atomic snapshot. New data is ingested, and all previous data from this plugin not present in the current batch is removed.
+
+## Multi-plugin properties
+
+Multiple plugins can independently report the **same node or relation** (identified by identical `NodeID` or relation `kind`/`from`/`to` triple). The catalog merges their data rather than overwriting it.
+
+**How it works:**
+
+- Each plugin's properties are stored under that plugin's name as a namespace — the catalog never mixes or overwrites another plugin's data.
+- The API returns `props` as an ordered list of `{ plugin, entries }` objects, one per contributing plugin:
```
akavel-reply: This has now changed IIUC (also needs fixing below in the JSON example)


```suggestion
- The API returns `pluginClaims` as an ordered list of `{ plugin, props }` objects, one per contributing plugin:
```

akavel-reply (changes requested):

docs/plugin-authoring.md
libera13: Thanks!

catalog/internal/httpapi/router_test.go
akavel-reply, line 153: There's a filter here, but there's only one relation in the data anyway, so this test looks rather misleading to me.

A possible suggestion I'd have would be to split this into 2 tests:
1. One test with no filter at all;
2. For second test, keep the filter, but add another relation in the sample data, so that we could see it gets filtered out.

Or maybe we could keep just the variant 2 - as you prefer.

libera13: Thanks for the suggestions. I added one relation to sample data and changed the test name to something more descriptive

akavel-reply (approved):

libera13:
https://github.com/naira-project/naira/pull/44

PR #75, opened 2026-06-22
Feat: run plugins as gRPC sidecars

### LINKED ISSUE #41: Implement Plugin Manifest
plugins manifest -> plugins_registry.json (goal: move plugins to separate binaries)

This requires an ADR for why we go with Go-Plugins.

### LINKED ISSUE #67: Catalog plugins should use go-plugin
As a result of [ADR](https://github.com/naira-project/architecture/blob/main/adr/004-backend-plugin-system.md) we need to implement plugins with using of go-plugin

Acceptance criteria:
- each plugins is a separate binary
- each plugin have separate container image
- plugin config should be used, something like plugins_registry.json [rfc](https://github.com/naira-project/architecture/blob/main/rfc/rfc-004-core-api-v1-design.md#plugins_registryjson) 
- contract between catalog and plugins, should be in .proto files

Out of scope (to do in other tickets)
- adding/removing plugins should not restart catalog

libera13:
## Description

This change separates plugins from the catalog by moving them into independent binaries. Implemented gRPC communication using Protobuf to bridge the two. Given that they share the same pod/network deployment, no additional authorization logic is required at this time.
 
Other notes:
- The catalog now retrieves the plugin name from PLUGIN_ADDRESSES, so the Name() method has been removed from the plugin interface.
example of PLUGIN_ADDRESSES:
value: "litellm=localhost:50051,mlflow=localhost:50052,depl-calls-svc=localhost:50053"
- depl-calls-svc is enabled in the kind platform and scans the local kind cluster where it is deployed (verified via local testing).
- Added `buf` tool to manage proto files.
- Added a wrapper for plugin initialization, which includes the gRPC server and Mermaid output.

ticket https://github.com/naira-project/naira/issues/41
              
## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [x] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

## AI Usage
gemini-3.5-flash for coding

## Screenshots 
One pod, four containers (catalog + three plugins)
Screenshot 2026-06-22 at 08 44 47


rootlsz (approved):
looks good 👏

catalog/cmd/catalog/main.go
```diff
@@ -20,21 +23,41 @@ func main() {
...
+		logger.Fatalf("failed to register plugins: %v", err)
+	}
+	defer cleanup()
+
+	service := catalog.NewService(catalog.NewMemoryStore(), logger, registeredPlugins)
 	router := httpapi.NewRouter(service, logger)
 	server := &http.Server{
 		Addr:              fmt.Sprintf(":%d", config.Port),
 		Handler:           router,
 		ReadHeaderTimeout: 5 * time.Second,
```
akavel-reply: I'd suggest adding:

```suggestion
		ReadHeaderTimeout: 5 * time.Second, // TODO: make configurable
```

catalog/internal/pluginmanager/plugins.go
```diff
@@ -0,0 +1,92 @@
...
+			cleanup()
+		}
+	}
+
+	return registered, cleanupAll, nil
+}
+
+// ConnectPlugin establishes a gRPC connection to the specified address.
+// It includes a retry mechanism that waits up to 10 seconds for the
+// connection to reach the READY state before timing out.
```
akavel-reply, line 50: I'd suggest adding:


```suggestion
// connection to reach the READY state before timing out.
//
// TODO: make timeout configurable
```

catalog/internal/pluginmanager/plugins.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	conn, err := grpc.NewClient(
+		address,
+		// Insecure is safe for now, plugins run as sidecars within the same pod
+		// and share the localhost network namespace. Upgrade to mTLS/Service Mesh
+		// if plugins are moved to separate pods.
```
akavel-reply: Hm, I think safe or not safe depends (as always?) on threat modeling in particular use case... I must admit I don't fully understand how things work in this situation. Is it possible you could provide some link with an overview/explanation where I could learn more and get better understanding of resistance of this setup to various attack vectors? Ideally as a link in the comment in code. Also, I'd like to propose still adding a TODO or FIXME here - even on localhost someone might want to still use encrypted connections for "defense in depth"; I understand we don't need it yet at our current phase of this project, but just to slightly bump up visibility of this place for future fixing.

catalog/pluginapi/proto/plugin/v1/plugin.proto
```diff
@@ -0,0 +1,31 @@
...
+  NodeID to = 3;
+  map<string, string> properties = 4;
+}
+
+message CollectResponse {
+  repeated NodeClaim nodes = 1;
+  repeated RelationClaim relations = 2;
+}
+
+message CollectRequest {}
```
akavel-reply: (nit) Please put CollectRequest first, before CollectResponse - i.e. in chronological order, this will make them easier to read

plugins/pkg/pluginapi/proto/buf.gen.yaml
akavel-reply, line 1: I see mentions of some "buf" tool, but it's not clear to me where can I get it, and also where I find its documentation.

catalog/pluginapi/proto/buf.gen.yaml
```diff
@@ -0,0 +1,8 @@
+version: v2
+plugins:
+  - remote: buf.build/protocolbuffers/go
```
akavel-reply: I'd like the versions of protoc-gen-go etc. generators to be pinned; does this tool support that?

plugins/pkg/api/grpc.go
```diff
@@ -0,0 +1,37 @@
...
+func (s *GRPCServer) Collect(ctx context.Context, _ *pluginv1.CollectRequest) (*pluginv1.CollectResponse, error) {
+	req, err := s.Impl.Collect(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("collecting from plugin implementation: %w", err)
+	}
+	return toProtoCollectResponse(req), nil
+}
+
+type GRPCClient struct {
+	client pluginv1.CatalogPluginServiceClient
```
akavel-reply: the Server has a public field, while the Client has a private field + a constructor; is there a reason for this difference? can we make the Client's field public and drop the constructor?

catalog/pluginapi/mapping.go
```diff
@@ -0,0 +1,105 @@
...
+		return NodeID{}
+	}
+	return NodeID{
+		Kind: p.Kind,
+		Path: p.Path,
+	}
+}
+
+func toProtoPropertyMap(properties PropertyMap) map[string]string {
+	if properties == nil {
```
akavel-reply: I think we don't need this check (also same in next func)

catalog/Dockerfile
```diff
@@ -1,18 +1,25 @@
...
-COPY pluginapi ./pluginapi
+RUN go mod download
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o /out/catalog ./cmd/catalog
+COPY catalog/internal/ ./catalog/internal/
+COPY catalog/pluginapi/ ./catalog/pluginapi/
+COPY catalog/cmd/catalog/ ./catalog/cmd/catalog/
 
-FROM gcr.io/distroless/static-debian12
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
```
akavel-reply: in this and other dockerfiles:
```suggestion
RUN CGO_ENABLED=0 \
```

catalog/Dockerfile
```diff
@@ -1,18 +1,25 @@
...
+    go build -ldflags="-w -s" -o /out/catalog ./catalog/cmd/catalog
```
akavel-reply: in this and other dockerfiles: why `-w -s`?

deploy/dev/infra/kubernetes/catalog.yaml
```diff
@@ -28,6 +28,31 @@ rules:
...
+kind: ClusterRole
+metadata:
+  name: catalog-depl-calls
+rules:
+  - apiGroups: [""]
+    resources: ["namespaces", "services"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["apps"]
+    resources: ["deployments"]
+    verbs: ["get", "list", "watch"]
```
akavel-reply: Do we need "watch" perms as of now?

deploy/dev/infra/kubernetes/catalog.yaml
```diff
@@ -28,6 +28,31 @@ rules:
     verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: catalog-depl-calls
```
akavel-reply: I'm confused by this name (also same in the next resource) - why is it like this?

deploy/dev/Taskfile.yml
```diff
@@ -25,6 +25,12 @@ vars:
   POSTGRES_MANIFEST: infra/kubernetes/postgres.yaml
   CATALOG_DIR: ../../catalog
   CATALOG_IMAGE: catalog:local
+  MLFLOW_PLUGIN_DIR: ../../plugins/cmd/mlflow
+  MLFLOW_PLUGIN_IMAGE: mlflow:local
+  LITELLM_PLUGIN_DIR: ../../plugins/cmd/litellm
+  LITELLM_PLUGIN_IMAGE: litellm:local
+  DEPL_CALLS_SVC_PLUGIN_DIR: ../../plugins/cmd/depl_calls_svc
```
akavel-reply: (optional) can we define `../../plugins/cmd` as some helper var, and use it here and above?

deploy/dev/Taskfile.yml
```diff
@@ -154,7 +161,43 @@ tasks:
...
+
+  mlflow:image:build:
+    desc: Build the mlflow Docker image
+    cmds:
+      - docker build -t {{.MLFLOW_PLUGIN_IMAGE}} -f {{.MLFLOW_PLUGIN_DIR}}/Dockerfile ../../
+
+  litellm:image:build:
+    desc: Build the litellm Docker image
+    cmds:
+      - docker build -t {{.LITELLM_PLUGIN_IMAGE}} -f {{.LITELLM_PLUGIN_DIR}}/Dockerfile ../../
```
akavel-reply: This cmd is very similar to the one above, and also one or more below; does Taskfile allow to create some macros or functions or something, to reduce repetitiveness?

plugins/cmd/depl_calls_svc/main.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"log"
+	"os"
+
+	"github.com/naira-project/naira/plugins/internal/pluginmain"
+	"go-simpler.org/env"
+)
+
+const defaultPort = 50053
```
akavel-reply: I'm not convinced; I'd rather suggest having all plugins use the same default port, set in pluginmain.Run. Currently this suggests each plugin must be aware of all previously plugins ever created, and also if I run a plugin standalone I can't easily always test them by connecting to the same port.

plugins/cmd/depl_calls_svc/main.go
```diff
@@ -0,0 +1,28 @@
...
+
+type pluginConfig struct {
```
akavel-reply:
```suggestion
type config struct {
```

plugins/cmd/depl_calls_svc/main.go
```diff
@@ -0,0 +1,28 @@
...
+
+func main() {
+	var raw pluginConfig
+	if err := env.Load(&raw, nil); err != nil {
+		log.Fatalf("failed to load depl_calls_svc config: %v", err)
+	}
+
+	logger := log.New(os.Stdout, "", log.LstdFlags)
+
+	impl := New(raw)
```
akavel-reply: Maybe we could put all that in pluginmain.Run as well?

Instead of `impl := New(raw)`, we could do: `impl := cfg.Build()` or something like that.

plugins/cmd/depl_calls_svc/plugin.go
```diff
@@ -20,8 +20,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 
-	"github.com/naira-project/naira/catalog/internal/kubeconn"
 	"github.com/naira-project/naira/catalog/pluginapi"
+	"github.com/naira-project/naira/plugins/internal/kubeconn"
```
akavel-reply: If plugins are now in `$REPO/plugins/cmd/...`, it seems to me like a directory with their shared utils would better fit in: `$REPO/plugins/internal/...`

plugins/cmd/depl_calls_svc/plugin.go
akavel-reply, line 1: Maybe we could merge this file into `main.go`? The `func main()` will potentially end up being very tiny

plugins/cmd/litellm/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+	logger := log.New(os.Stdout, "", log.LstdFlags)
+
+	httpClient := &http.Client{
+		Timeout: raw.HTTPTimeout,
+	}
+
+	impl := New(httpClient, logger, pluginConfig{
+		BaseURL: strings.TrimSpace(raw.BaseURL),
+		APIKey:  strings.TrimSpace(raw.APIKey),
```
akavel-reply: I'd skip the TrimSpace

plugins/cmd/litellm/plugin.go
```diff
@@ -29,33 +29,23 @@ const (
...
-}
-
-func New(httpClient *http.Client, logger *log.Logger, config Config) *Plugin {
+func New(httpClient *http.Client, logger *log.Logger, config pluginConfig) *Plugin {
 	return &Plugin{
 		httpClient:          httpClient,
 		logger:              logger,
 		config:              config,
-		appIdentityProvider: newAppIdentityProvider(logger),
+		appIdentityProvider: nil,
```
akavel-reply: do we need this field anymore? I don't immediately see where it's being set, am I missing it?

plugins/cmd/mlflow/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+	httpClient := &http.Client{
+		Timeout: raw.HTTPTimeout,
+	}
+
+	logger := log.New(os.Stdout, "", log.LstdFlags)
+
+	impl := New(httpClient, pluginConfig{
+		BaseURL:     strings.TrimSpace(raw.BaseURL),
+		BearerToken: strings.TrimSpace(raw.BearerToken),
```
akavel-reply: I'd skip the TrimSpaces

plugins/internal/pluginmain/pluginmain.go
```diff
@@ -0,0 +1,80 @@
...
+	pluginv1.RegisterCatalogPluginServiceServer(s, &pluginapi.GRPCServer{Impl: p})
+
+	logger.Printf("Plugin listening via gRPC on %v", lis.Addr())
+	if err := s.Serve(lis); err != nil {
+		logger.Fatalf("failed to serve gRPC: %v", err)
+	}
+}
+
+func getPort(defaultPort int, logger *log.Logger) int {
+	envPort := os.Getenv(portEnv)
```
akavel-reply: As I wrote elsewhere, I'd suggest keeping the same default port across all plugins; then, I think we could also remove this whole func and just use `go-simpler.org/env` library instead

akavel-reply (commented):
I haven't checked tests yet.

plugins/pkg/pluginapi/proto/buf.gen.yaml
libera13: If we agree to use mise, then mise will "get it". Docs https://buf.build/docs/cli/#why-the-buf-cli

catalog/pluginapi/proto/buf.gen.yaml
libera13: Yes, I will add it

plugins/pkg/api/grpc.go
libera13: The `client` object is used later in the code, so theoretically, someone could set it to `nil` if we make it public. The `server` object is passed in solely as part of the configuration, so it cannot be modified from within our code

Let's say it makes sense :), what do you think?

catalog/Dockerfile
libera13: correct, good for prod, not for dev phase

catalog/cmd/catalog/main.go
libera13: I'll implement it right away

catalog/internal/pluginmanager/plugins.go
libera13: I'll implement it right away

deploy/dev/infra/kubernetes/catalog.yaml
libera13: No :) removed

deploy/dev/infra/kubernetes/catalog.yaml
libera13: Good point, good to separate it more if we separate plugins

plugins/cmd/litellm/main.go
```diff
@@ -1,40 +1,272 @@
...
+}
+
 func main() {
-	var raw pluginConfig
-	if err := env.Load(&raw, nil); err != nil {
-		log.Fatalf("failed to load litellm config: %v", err)
+	app := pluginmain.New[config]()
+	cfg := app.Config()
+
+	httpClient := &http.Client{Timeout: cfg.HTTPTimeout}
```
akavel-reply: Why isn't this created inside `New()`?

plugins/pkg/pluginmain/pluginmain.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"net"
+	"os"
+
+	pluginapi "github.com/naira-project/naira/plugins/pkg/api"
+	pluginv1 "github.com/naira-project/naira/plugins/pkg/api/proto/plugin/v1"
+	"go-simpler.org/env"
```
akavel-reply, line 13: Same as current package's directory name is matching the `pluginmain` package name, I'd suggest to also rename the other directories imported here, e.g. so that above becomes:


```suggestion
	"github.com/naira-project/naira/plugins/pkg/pluginapi"
	"github.com/naira-project/naira/plugins/pkg/pluginapi/proto/pluginv1"
```

plugins/pkg/pluginmain/pluginmain.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+type serverConfig struct {
+	Port int `env:"PORT" default:"50051"`
+}
+
+type App[C any] struct {
+	pluginConfig C
+	serverConfig serverConfig
+	logger       *log.Logger
```
akavel-reply: I'd suggest to make at least pluginConfig & logger exported, and drop the getters:


```suggestion
	PluginConfig C
	Logger       *log.Logger
	
	serverConfig serverConfig
```

plugins/pkg/pluginmain/pluginmain.go
```diff
@@ -0,0 +1,104 @@
...
+	if *mermaidFlag {
+		res, err := p.Collect(context.Background())
+		if err != nil {
+			a.logger.Fatalf("failed to collect: %v", err)
+		}
+		printMermaid(res, a.logger)
+		return
+	}
+
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", a.serverConfig.Port))
```
akavel-reply: Per our recent discussion, AFAIU we want to listen only on 127.0.0.1 here

plugins/cmd/mlflow/main.go
```diff
@@ -21,32 +23,38 @@ const (
...
 	}
 }
 
-func (*Plugin) Name() string {
-	return pluginName
+func main() {
+	app := pluginmain.New[config]()
+	cfg := app.Config()
+
+	httpClient := &http.Client{Timeout: cfg.HTTPTimeout}
```
akavel-reply: can we move this into New()?

catalog/cmd/catalog/main.go
```diff
@@ -20,21 +23,41 @@ func main() {
...
+
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			logger.Fatalf("server failed: %v", err)
+		}
+	}()
+
+	<-ctx.Done()
+	logger.Println("shutting down catalog service...")
+
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
akavel-reply: another hardcoded timeout, I think should be configurable

catalog/internal/pluginmanager/plugins.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Register connects to each plugin sidecar by its configured name and gRPC
+// address and returns the registered plugins keyed by plugin name.
+func Register(plugins map[string]string, logger *log.Logger, timeout time.Duration) (map[string]pluginapi.Plugin, func(), error) {
+	registered := make(map[string]pluginapi.Plugin, len(plugins))
+	var cleanups []func()
+
+	for name, addr := range plugins {
+		if name == "" || addr == "" {
+			continue
```
akavel-reply: this is weird, should we rather at least warn, or maybe even return error?

plugins/pkg/api/grpc.go
akavel-reply: Well, FWIW, someone can also:

1. pass a `nil` as argument to `NewGRPCClient()`
2. (I think) set the field to `nil` by assigning a zero struct: `mygrpcclient = GRPCClient{}`

while making the field exported would let us remove the constructor - same as GRPCServer has no constructor.

As to "within our code", I suspect what you're trying to say, but I don't necessarily agree with it, and so I'd like to challenge it: what do you mean by "our"?

- from one perspective, all the code in this repository is "our", so there's no difference;
- from perspective of `pluginapi` package, both GRPCServer and GRPCClient are used outside this package, so someone can as well change the `Impl` field and "set it to `nil`". From this perspective, I can only see a confusing assymetry between those two structs. And, honestly speaking, I like looking at packages from their isolated perspective, assuming zero knowledge about how they're being used by some faraway code - because such faraway code will change in the future and consequences those changes won't trickle down back to this code here.

Based on the last bullet point above, I'd like to ask to unify the approach between those two structs - either expose the field in both cases, or make it non-exposed in both cases.

akavel-reply (approved):

PR #80, opened 2026-06-23
Feat: mise config file

### LINKED ISSUE #77: Mise/Devcontainers to have the same tools across team
It is worth enforcing that we have the same versions of tools installed. For example, the same version of buf, kind, etc.

Consider using Devcontainers or/and mise

Two goals:
- new developer can quickly install all needed tools to run/develop Naira
- team working on Naira is using the same versions of tools

libera13:
## Description

Adding mise config file. Useful for ensuring that contributors have the same versions of tools installed
In addition I pinned version of k8s cluster, so it can match kubectl version

ticket https://github.com/naira-project/naira/issues/77

### DIFF
--- a/deploy/dev/infra/kind/kind-config.yaml
+++ b/deploy/dev/infra/kind/kind-config.yaml
@@ -4,3 +4,4 @@ name: naira-idp
 
 nodes:
   - role: control-plane
+    image: kindest/node:v1.36.1 # update kubectl in mise.toml if changing version 
--- a/mise.toml
+++ b/mise.toml
@@ -1,8 +1,9 @@
 [tools]
-go = "1.26"
-python = "3.11"
-kind = "0.32"
-kubectl = "1.36"
-helm = "4.2"
 buf = "1.71"
+go = "1.26"
 task = "3.51"
+# deploy/
+helm = "4.2"
+kubectl = "1.36" # update kind in kind-config.yaml if changing version 
+kind = "0.32"
+python = "3.11"
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,8 @@
+[tools]
+go = "1.26"
+python = "3.11"
+kind = "0.32"
+kubectl = "1.36"
+helm = "4.2"
+buf = "1.71"
+task = "3.51"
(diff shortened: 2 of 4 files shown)


mise.toml
akavel-reply, line 1: Please add a basic intro in the Readme how to start with mise for development.

mise.toml
```diff
@@ -0,0 +1,8 @@
+[tools]
+go = "1.26"
+python = "3.11"
+kind = "0.32"
+kubectl = "1.36"
```
akavel-reply: is there some way in mise to split the tools needed for various usecases? for example, when developing, I might not need python or helm

mise.toml
libera13: @mkorbi what do you think about it? Can I add note about it to Readme in root, or maybe you want to keep it without such info? I'm not sure if you have some plan for Readme

mkorbi: I don't know, if we are all agree to use it, we can spend a "How to start Develop" section or so in the readme, and later in the docs.

mise.toml
libera13: Yes, there is this option. In different files e.g. mise.ci.toml we can keep different tools - https://mise.jdx.dev/configuration/environments.html#config-environments

However I'm not sure if it's worth to split them at this stage, we can talk about it

akavel-reply: If not splitting, could we at least group the tools for now (with explanatory comments), and in the groups sort them alphabetically?

For example, if I won't be deploying `deploy/dev/`, I don't need Python AFAIU. E.g. something more or less (please double-check, I'm only sketching the idea) like:
```toml
[tools]
buf = "1.71"
go = "1.26"
task = "3.51"
# deploy/
helm = ...
# deploy/dev/
kind = ...
kubectl = ...
python = ...

File: deploy/dev/infra/kind/kind-config.yaml [hunk 2]
```diff
@@ -3,4 +3,5 @@ apiVersion: kind.x-k8s.io/v1alpha4
 name: naira-idp
 
 nodes:
-  - role: control-plane
\ No newline at end of file
+  - role: control-plane
+    image: kindest/node:v1.36.1 # update kubectl in mise.toml if changing version
```
akavel-reply, line 7: Should we also add a matching comment in `mise.toml`?

akavel-reply (approved):

PR #83, opened 2026-06-29
feat: plugin scanning LiteLLM models from Deployment Secrets

akavel-reply:
## Description

Plugin `depl_uses_litellm` scans k8s Deployments for LiteLLM API keys stored in
referenced Secrets, then queries each configured LiteLLM host to discover
which models that key can access, and emits deployment→model "uses_model"
Relations.

Issue: https://github.com/naira-project/product/issues/43

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [x] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

## AI Usage

Claude Code with sonnet-4.6 for base code, then lots of tweaks and refactorings on top.

When run on a test setup on kind, I'm getting the following result (hostname adjusted):

```console
$ DEPL_USES_LITELLM_HOSTS=litellm.example.com go run ./plugins/cmd/depl_uses_litellm/  -mermaid
graph TD
    n0["deployment/bd44ffd2-7342-4441-b048-c661c8c11b95/default/chat2"]
    n1["model/litellm/litellm.example.com/mistral-small-4"]
    n0 -->|"uses_model"| n1
```

which, as a mermaid graph, gives the following:
```mermaid
graph TD
    n0["deployment/bd44ffd2-7342-4441-b048-c661c8c11b95/default/chat2"]
    n1["model/litellm/litellm.example.com/mistral-small-4"]
    n0 -->|"uses_model"| n1
```


plugins/cmd/depl_uses_litellm/plugin.go
```diff
@@ -0,0 +1,334 @@
...
+		relations     []pluginapi.RelationClaim
+		seenRelations = make(map[fromTo]struct{})
+	)
+
+	for _, d := range deplsWithAPIKeys {
+		deplID := pluginapi.NodeID{
+			Kind: pluginapi.NodeKindDeployment,
+			Path: clusterID + "/" + d.namespace + "/" + d.deployment,
+		}
+		deplsByPath[deplID.Path] = pluginapi.NodeClaim{ID: deplID}
```
libera13, line 150: here we will publish all deployments with secrets that match the regex. Probably we might match keys that do not belong to Litellm. While I am not suggesting that we should only publish deployments after verifying connection with litellm, but I start to wonder if maybe we should start to keep small readme per plugin, which describes what a user installing plugin can expect

plugins/cmd/depl_uses_litellm/plugin.go
```diff
@@ -0,0 +1,334 @@
...
+		for _, depl := range depls.Items {
+			ns, name := depl.GetNamespace(), depl.GetName()
+			seenSecretValues := make(map[string]struct{})
+
+			for secretName := range referencedSecretsNames(depl.Object) {
+				nameWithNs := nameWithNs{name: secretName, ns: ns}
+				if _, seen := seenSecrets[nameWithNs]; seen {
+					continue
+				}
+				seenSecrets[nameWithNs] = struct{}{}
```
libera13: Gemini finding (looks valid): 
If you have two deployments in the same namespace referencing the same secret, then second deployment will be skipped

plugins/cmd/depl_uses_litellm/plugin.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+	}
+	return names
+}
+
+// fetchModels calls GET <baseURL>/v1/models with the given key and returns a list of model IDs.
+// A 401/403 response means the key is not valid for that host and empty results are returned.
+func fetchModels(client *http.Client, baseURL, apiKey string) ([]string, error) {
+	addr := baseURL + "/v1/models"
+	req, err := http.NewRequest(http.MethodGet, addr, nil)
```
libera13: what do you think about using`NewRequestWithContext` to use context from catalog?

plugins/cmd/depl_uses_litellm/plugin.go
```diff
@@ -0,0 +1,334 @@
...
+		if _, seen := keyToHostsToModels[d.secret]; seen {
+			continue
+		}
+		hostToModels := make(map[string][]string)
+		for _, host := range p.config.Hosts {
+			models, err := fetchModels(p.httpClient, "https://"+host, d.secret)
+			if err != nil {
+				log.Printf("%s: WARN: fetching models from %s (key ...%s): %v",
+					pluginName, host, d.secret[len(d.secret)-4:], err)
+				if !errors.Is(err, http.ErrSchemeMismatch) {
```
libera13: gemini finding (I haven't checked it):
The http.ErrSchemeMismatch error is returned by http.Client only in specific situations involving redirects. If you try to make an HTTPS request to a port that only supports HTTP, you will receive a TLS network error (e.g., “tls: first record does not look like a TLS handshake”), not an ErrSchemeMismatch error.

plugins/cmd/depl_uses_litellm/plugin.go
```diff
@@ -0,0 +1,334 @@
...
+			}
+		}
+		keyToHostsToModels[d.secret] = hostToModels
+	}
+
+	// Build Node & Relation claims.
+	type fromTo struct{ from, to pluginapi.NodeID }
+	var (
+		deplsByPath   = make(map[string]pluginapi.NodeClaim) // key: "clusterID/namespace/name"
+		modelsByPath  = make(map[string]pluginapi.NodeClaim) // key: "host/modelID"
```
libera13, line 140: Just a thought. From what I understand if someone will change exposed port in litellm Service, it will change the nodes paths, strange a little bit. But anyway I didn't figure out better identificator of litellm instance for now. Maybe in config, beside the hostname, require also the litellm name (e.g. litellm-prod), maybe

plugins/cmd/depl_uses_litellm/plugin_test.go
```diff
@@ -0,0 +1,492 @@
...
+	"k8s.io/client-go/dynamic/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+const (
+	testClusterID = "test-cluster-uid-1234"
+	testAPIKey    = "sk-123456789_123456789_12" // should match default config.APIKeyRegexp
+)
+
+var defaultKeyRegexp = compileDefaultAPIKeyRegexp()
```
libera13: not used

plugins/cmd/depl_uses_litellm/plugin_test.go
```diff
@@ -0,0 +1,492 @@
...
+				secret("team-a", "secret-4", stringmap{
+					"FFF": "sk-FFF",
+				}),
+				secret("team-a", "secret-5", stringmap{
+					"GGG": "sk-GGG",
+				}),
+			},
+			want: []deploymentWithSecret{
+				{namespace: "team-a", deployment: "app", secret: "sk-AAA"},
+				{namespace: "team-a", deployment: "app", secret: "sk-BAD"}, // FIXME: should not be here?
```
libera13, line 78: Is it important to fix this now?

plugins/cmd/depl_uses_litellm/plugin.go
akavel-reply: In the end, thanks to your other idea for names in config, I was able to remove it and instead require explicit protocol in the URL in config.

plugins/cmd/depl_uses_litellm/plugin_test.go
akavel-reply: It would be annoyingly complicated to do, and the benefit at current PoC stage is IMO not worth it; for now, I added it to a "Known Issues" section in the readme instead. Thanks for bringing it up!

plugins/cmd/depl_uses_litellm/README.md
```diff
@@ -0,0 +1,22 @@
+# depl_uses_litellm plugin
+
+depl\_uses\_litellm scans k8s Deployments for LiteLLM API keys stored in referenced Secrets, then queries each configured LiteLLM host to discover which models that key can access, and emits deployment→model "uses\_model" relations.
+
+## Known Issues {#hdr-Known_Issues}
```
libera13: Is this {#hdr-Known_Issues} needed? Or can be removed?

mise.toml
```diff
@@ -1,6 +1,10 @@
+[alias]
```
libera13, line 1: When I was installing it, I got 
`mise WARN  deprecated [alias]: [alias] is deprecated, use [tool_alias] instead in ~/Documents/github/reply/naira/mise.toml         `

```suggestion
[tool_alias]
```

akavel-reply: I'd like to keep it as is for now. My version of mise is older and doesn't support `[tool_alias]` yet. I'll have to look into upgrading it, but in meantime, although this is a warning, it still works with your version, right?

libera13: Fine for me, yes it works

libera13 (approved):

PR #90, opened 2026-07-07
OpenMetadata plugin

Daviidg:
## Description

This PR contains the implementation of the OpenMetadata plugin. The plugin ingests tables present in an OpenMetadata instance into Naira and displays them in a newly implemented Dataset Catalog UI. The code additionally contains the Helm and yaml configs necessary to deploy a simple OpenMetadata instance locally as well as a seed script to populate it with some data.

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 📝 Documentation update
- [ ] 🔧 Refactoring (no functional changes)
- [x] 🏗️ Infrastructure / CI / Helm

## How Has This Been Tested?

- [ ] `go test ./...` passes
- [ ] `helm lint` passes
- [ ] Manual testing via Docker Compose
- [x] Manual testing on Kubernetes

## Checklist

- [x] My code follows the project's coding standards ([AGENTS.md](AGENTS.md))
- [ ] I have added tests that prove my fix/feature works
- [x] New and existing unit tests pass locally
- [ ] I have updated the documentation (if applicable)
- [ ] My changes generate no new warnings

Issue #7, opened 2026-05-05
Plugin Lifecycle Engine


mkorbi:
## 📋 Overview

**User Story:** As **Platform Peter** (Platform Engineer), I want a standardized engine within Naira to install, configure, update, and uninstall platform plugins, so that I can extend the platform's capabilities (like adding a new AI Gateway or GitOps engine) without modifying core platform code or requiring a full system redeployment.

---

## 🎯 Context
To achieve a vendor-agnostic architecture and prevent lock-in, Naira's core must remain lightweight. All integrations with specific tools (e.g., LiteLLM, ArgoCD, LGTM) will be handled as plugins. The Core Plugin Lifecycle Engine is the control plane component responsible for orchestrating the state of these plugins within a workspace or globally.

---

### 1. Extension API Contracts
Strictly defined interfaces (e.g., gRPC APIs or Kubernetes CRDs) that plugins must implement to hook into the platform. Extension points include:
- **UI Extensions:** Registering micro-frontends with OpenMFP.
- **Data Extensions:** Pushing metadata/aspects into the Context Engine.
- **Action Extensions:** Exposing custom operations to the platform.

### 2. Lifecycle Orchestrator
A controller that manages the state machine of a plugin instance:
- **Installation:** Deploying the plugin controller/operator (e.g., via Helm or KRM).
- **Configuration:** Injecting workspace-specific configuration (credentials, endpoints) securely.
- **Updates:** Managing version rollouts safely.
- **Uninstallation:** Clean teardown of plugin resources and removal of registered extensions.

### 3. Health & Status Monitoring
Continuous monitoring of installed plugins to ensure they are responsive and correctly synced with the core Context Engine, reporting status back to the platform UI.

---

## 🛠 Functional Requirements

- **API/CRD Definition:** Define clear API contracts for plugins to register their capabilities (UI, Data, Action).
- **Lifecycle Management Operations:** Provide endpoints/mechanisms to Install, Configure, Update, Suspend, and Uninstall plugins.
- **Status Reporting:** Expose the health and sync status of all installed plugins.
- **Workspace Isolation:** Ensure plugins installed within a specific KCP workspace only operate within that workspace's context unless explicitly installed globally.

---

## 🛡️ Non-Functional Requirements (NFRs)

- **Resilience:** The crash or failure of a single plugin must not impact the core Naira platform or other installed plugins.
- **Backward Compatibility:** Core platform upgrades must preserve compatibility with existing plugins adhering to the defined API contracts.
[...]

PR #93, opened 2026-07-08
feat: openmetadata plugin 

### LINKED ISSUE #10: Reference Integration: Data Catalog (OpenMetadata)
## 📋 Overview

**User Story:** As an **AI Engineer rAIner**, I want to discover datasets used for training models, and as **Platform Peter**, I want to ensure these datasets are properly governed and tracked. We want to use OpenMetadata as the source of truth, and have that data reflected in Naira.

---

## 🎯 Context
Data is foundational to AI. Naira needs a dataset registry. OpenMetadata is a leading open-source data catalog that provides robust lineage and governance features. This plugin integrates OpenMetadata, pulling dataset information and, crucially, data-to-model lineage into the Naira Context Engine.

---

### 1. OpenMetadata Sync Controller
A plugin component that connects to the OpenMetadata API to ingest dataset metadata, schemas, and lineage information.

### 2. Lineage Resolution
The plugin translates OpenMetadata's lineage graph into Naira's Relationship model (e.g., mapping an OpenMetadata `Table` to a Naira `Dataset` entity, and establishing `trainedOn` relationships with Naira `Model` entities if that lineage exists in OMD).

### 3. OpenMFP UI Extension
The plugin provides custom UI components to render data schemas and quality metrics sourced from OpenMetadata directly within the Naira dataset view.

---

## 🛠 Functional Requirements

- **Connection Configuration:** Configure connection details (URL, JWT token) for the OpenMetadata instance.
- **Dataset Sync:** Sync Table/Dataset entities, their schemas, descriptions, and tags into Naira.
- **Lineage Sync:** If OpenMetadata tracks lineage between datasets and ML models, sync those relationships into the Naira Context Engine.
- **Custom UI:** Display data schema and quality data in the Naira UI.

---

## ✅ Acceptance Criteria
- [ ] The OpenMetadata plugin is installed and configured.
- [ ] A dataset registered in OpenMetadata appears in the Naira Asset Catalog.
- [ ] The schema of the dataset is viewable in Naira via an injected MFE.
- [ ] If lineage exists in OpenMetadata linking a dataset to a model, that relationship is queryable in Naira's Context Engine.

Daviidg:
## Description

This PR contains the implementation of the OpenMetadata plugin. The plugin ingests tables present in an OpenMetadata instance into Naira. The code additionally contains the Helm and yaml configs necessary to deploy a simple OpenMetadata instance locally as well as a seed script to populate it with some data.

It is the same as this [PR #90 ](https://github.com/naira-project/naira/pull/90), just without the UI commit.

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [ ] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist


plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+	endpoint := strings.TrimRight(p.config.BaseURL, "/") + "/api/v1/users/login"
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, bytes.NewReader(body))
+	if err != nil {
+		return "", fmt.Errorf("building OpenMetadata login request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	var payload loginResponse
+	if err := p.doJSON(req, "login", &payload); err != nil {
+		return "", err
```
akavel-reply: (minor) We could avoid doing "naked `return err`" here, and also avoid the need to pass the "label" to `doJSON()` - for example through something like this:

```
    if err := p.doJSON(req, &payload); err != nil {
        return "", fmt.Errorf("executing request with JSON payload: %w")
    }
...
func (...) doJSON(...) ... {
    resp, err := p.httpClient.Do(req)
    if err != nil {
        return fmt.Errorf("executing HTTP request: %w")
    }
    ...
        return fmt.Errorf("got HTTP status %s: %s...", resp.Status, strings.TrimSpace(...))
    ...
        return fmt.Errorf("decoding response: %w", err)
    ...

File: plugins/cmd/openmetadata/main.go [hunk 2]
```diff
@@ -0,0 +1,437 @@
...
+	}
+
+	token := strings.TrimSpace(payload.AccessToken)
+	if token == "" {
+		return "", errors.New("OpenMetadata login response did not contain an access token")
```
akavel-reply: (minor) all the error messages in this function mention "OpenMetadata login"; the callsite then wraps them in a message saying: "authenticating with OpenMetadata: ...". This doubles the "OpenMetadata" name in the resulting message, and also effectively doubles a "login" meaning (the "authenticating" word in that case seems just a different way of saying "logging in"). It's enough to mention that information in one place. In the project, we're currently [choosing to mention such information in caller's error text, and not mention it in callee's](https://github.com/naira-project/naira/blob/787d79361ab7db67921103360994449a41bcbd7a/AGENTS.md#L22). I'd thus suggest to remove the text fragment `OpenMetadata login` from all error messages in this (`login()`) function.

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+	edges = append(edges, payload.DownstreamEdges...)
+
+	return edges, nil
+}
+
+// TODO: add pagination; tables beyond the first page are currently dropped.
+func (p *Plugin) fetchTables(ctx context.Context, token string) ([]table, error) {
+	endpoint, err := url.Parse(p.config.BaseURL + "/api/v1/tables")
+	if err != nil {
+		return nil, fmt.Errorf("building OpenMetadata tables URL: %w", err)
```
akavel-reply: (minor) see comments in `login()` - please remove "OpenMetadata tables" from error messages in this func.

```suggestion
		return nil, fmt.Errorf("building API URL: %w", err)
```

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("building OpenMetadata tables request: %w", err)
+	}
+	addAuthorization(req, token)
+
+	var payload tablesResponse
+	if err := p.doJSON(req, "tables", &payload); err != nil {
+		return nil, err
```
akavel-reply: (minor) please don't return "naked error" - see comments in `login()` func

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+		tables = append(tables, table{
+			ID:          strings.TrimSpace(item.ID),
+			Name:        strings.TrimSpace(item.Name),
+			FQN:         strings.TrimSpace(item.FullyQualifiedName),
+			Description: strings.TrimSpace(item.Description),
+			TableType:   strings.TrimSpace(item.TableType),
+			Platform:    strings.TrimSpace(item.ServiceType),
+			Columns:     item.Columns,
+			Tags:        collectTagFQNs(item.Tags),
+			Owners:      collectOwnerNames(item.Owners),
```
akavel-reply: Why we transform a `tableItem` struct into a new `table` struct like this? couldn't we just return `payload.Data`? I don't clearly see yet how this transformation adds enough value to justify having to maintain it.

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+
+	tables := make([]table, 0, len(payload.Data))
+	for _, item := range payload.Data {
+		tables = append(tables, table{
+			ID:          strings.TrimSpace(item.ID),
+			Name:        strings.TrimSpace(item.Name),
+			FQN:         strings.TrimSpace(item.FullyQualifiedName),
+			Description: strings.TrimSpace(item.Description),
+			TableType:   strings.TrimSpace(item.TableType),
+			Platform:    strings.TrimSpace(item.ServiceType),
```
akavel-reply: why we're changing the name of the field here?

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+		NativeType  string `json:"native_type,omitempty"`
+		Description string `json:"description,omitempty"`
+	}
+
+	payload := make([]columnProjection, 0, len(columns))
+	for _, item := range columns {
+		payload = append(payload, columnProjection{
+			Name:        strings.TrimSpace(item.Name),
+			Type:        strings.TrimSpace(item.DataType),
+			NativeType:  strings.TrimSpace(item.DataTypeDisplay),
```
akavel-reply: why we're changing the names of the fields here?

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+}
+
+func datasetPath(table table) string {
+	if table.FQN != "" {
+		return pluginName + "/" + table.FQN
+	}
+	if table.Name != "" {
+		return pluginName + "/" + table.Name
+	}
+	return pluginName + "/" + table.ID
```
akavel-reply: I'm thinking, instead of using `pluginName` for prefix, could it make sense to use a prefix configurable in the plugin's config? For example, if the user has multiple instances of this plugin running, configured for different OpenMetadata servers, maybe this prefix could then differentiate the resulting nodes? (It could maybe still default to `openmetadata` if a different value was not provided by user?)

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+			properties[propertyKeyOwners] = strings.Join(table.Owners, ",")
+		}
+		if columnsJSON := marshalColumns(table.Columns); columnsJSON != "" {
+			properties[propertyKeyColumnsJSON] = columnsJSON
+		}
+
+		nodeID := pluginapi.NodeID{Kind: pluginapi.NodeKindDataset, Path: datasetPath(table)}
+		nodes = append(nodes, pluginapi.NodeClaim{ID: nodeID, Properties: properties})
+		if table.ID != "" {
+			nodeByEntityID[table.ID] = nodeID
```
akavel-reply: why we're using only `table.ID` here, but in `datasetPath()` we're trying `FQN`, `Name`, and `ID`?

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+
+func (p *Plugin) tableURL(fqn string) string {
+	if strings.TrimSpace(fqn) == "" {
+		return ""
+	}
+
+	return strings.TrimRight(p.config.BaseURL, "/") + "/table/" + url.PathEscape(fqn)
+}
+
+func datasetPath(table table) string {
```
akavel-reply: Could you please add a `godoc` for this function, explaining the logic of choosing FQN, falling back to Name, falling back to ID? I don't know much about OpenMetadata, and it's quite nonobvious to me as such - I think it could be helpful to have some explanation here to help the reader of this code better understand the thought process and reasons for this.

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+	}
+
+	relations, err := p.collectLineage(ctx, tables, nodeByEntityID, token)
+	response := pluginapi.CollectResponse{Nodes: nodes, Relations: relations}
+	if err != nil {
+		return response, err
+	}
+
+	return response, nil
+}
```
akavel-reply: can be simplified to:

```suggestion
	return response, err
}
```

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+		}
+
+		nodeID := pluginapi.NodeID{Kind: pluginapi.NodeKindDataset, Path: datasetPath(table)}
+		nodes = append(nodes, pluginapi.NodeClaim{ID: nodeID, Properties: properties})
+		if table.ID != "" {
+			nodeByEntityID[table.ID] = nodeID
+		}
+	}
+
+	relations, err := p.collectLineage(ctx, tables, nodeByEntityID, token)
```
akavel-reply: (minor, style) If we're building the `relations` slice in a helper method, then for consistent approach to levels of abstraction in the code, I'd suggest to also create a sibling `collectTables()` (or other name) method, that would return `nodes`.

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+
+	endpoint := strings.TrimRight(p.config.BaseURL, "/") + "/api/v1/users/login"
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, bytes.NewReader(body))
+	if err != nil {
+		return "", fmt.Errorf("building OpenMetadata login request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	var payload loginResponse
+	if err := p.doJSON(req, "login", &payload); err != nil {
```
akavel-reply: (optional) FWIW, the `loginResponse` type doesn't seem used anywhere else from what I can see, so it could be even declared here anonymously - e.g. you could do it like so:

```suggestion
	var loginResponse struct {
		AccessToken string `json:"accessToken"`
	}
	if err := p.doJSON(req, "login", &loginResponse); err != nil {
```

plugins/cmd/openmetadata/main.go
```diff
@@ -0,0 +1,437 @@
...
+		return "", nil
+	}
+	if email == "" || password == "" {
+		return "", errors.New("both OPENMETADATA_ADMIN_EMAIL and OPENMETADATA_ADMIN_PASSWORD must be set, or neither")
+	}
+
+	body, err := json.Marshal(loginRequest{
+		Email:    email,
+		Password: base64.StdEncoding.EncodeToString([]byte(password)),
+	})
```
akavel-reply: (optional) FWIW, this could also be done without defining a type, in one of two ways:

```suggestion
	body, err := json.Marshal(map[string]string{
		"email":    email,
		"password": base64.StdEncoding.EncodeToString([]byte(password)),
	})
```
or:

```suggestion
	body, err := json.Marshal(struct{
		Email    string `json:"email"`
		Password string `json:"password"`
	}{
		Email:    email,
		Password: base64.StdEncoding.EncodeToString([]byte(password)),
	})
```

akavel-reply (commented):
generally looks good to me; just a bunch of smaller mostly stylistic/quality suggestions

plugins/cmd/openmetadata/main.go
Daviidg: The lineage API references entities exclusively by their ID. I am using FQN in datasetPath because its also unique but more readable. I could simplify the code and just use ID everywhere

plugins/cmd/openmetadata/main.go
Daviidg: Other plugins already use the pluginName/<identifier> format. I'll leave it as is for now, but I completely agree we should change this across all plugins down the line.

plugins/cmd/openmetadata/main.go
Daviidg: I went for option 1

plugins/cmd/openmetadata/main.go
Daviidg: simplified the code to just use the ID, datasetPath is no longer needed

plugins/cmd/openmetadata/main.go
akavel-reply: If I could ask you kindly to create a ticket for this cool idea for all plugins, I'd be grateful

plugins/cmd/openmetadata/main.go
akavel-reply: Sorry for coming back here and a bit revert: I actually now noticed we're returning naked error  here - I'd prefer we don't do that - OR write a comment explaining why we return naked error to justify an exception to the policy.

plugins/cmd/openmetadata/main.go
```diff
@@ -75,49 +75,47 @@ func (p *Plugin) Collect(ctx context.Context) (pluginapi.CollectResponse, error)
...
-			propertyKeyFQN:         table.FQN,
-			propertyKeySourceURL:   p.tableURL(table.FQN),
+			propertyKeyFQN:         table.FullyQualifiedName,
+			propertyKeySourceURL:   p.tableURL(table.FullyQualifiedName),
 		}
 
-		if table.Platform != "" {
-			properties[propertyKeyPlatform] = table.Platform
+		if table.ServiceType != "" {
+			properties[propertyKeyPlatform] = table.ServiceType
```
akavel-reply: still some renaming happening here

Daviidg: fixed now

plugins/cmd/openmetadata/main.go
Daviidg: Done: https://github.com/naira-project/naira/issues/7?issue=naira-project%7Cnaira%7C101

plugins/cmd/openmetadata/main.go
```diff
@@ -20,15 +20,15 @@ import (
...
-	propertyKeyTableType   = "table_type"
-	propertyKeyTags        = "tags"
-	propertyKeyOwners      = "owners"
-	propertyKeyColumnsJSON = "columns_json"
+	propertyKeySource             = "source"
+	propertyKeyDescription        = "description"
+	propertyKeyFullyQualifiedName = "fullyQualifiedName"
+	propertyKeySourceURL          = "source_url"
+	propertyKeyServiceType        = "serviceType"
+	propertyKeyTableType          = "tableType"
```
akavel-reply: Currently, we use `snake_case` for properties names in plugins; we might want to change that, but if yes, let's change in all of them; regardless, let's please stay consistent.

Daviidg: changed them to snake_case for now to match the other plugins

akavel-reply (approved):

deploy/dev/Taskfile.yml
```diff
@@ -168,6 +187,56 @@ tasks:
...
+      - helm repo add {{.OPENMETADATA_HELM_REPO_NAME}} {{.OPENMETADATA_HELM_REPO_URL}} --force-update
+      - helm repo update {{.OPENMETADATA_HELM_REPO_NAME}}
+      - kubectl create namespace {{.OPENMETADATA_NAMESPACE}} --dry-run=client -o yaml | kubectl apply -f -
+      - kubectl apply -f {{.OPENMETADATA_SECRETS_MANIFEST}}
+      - helm upgrade --install {{.OPENMETADATA_DEPS_RELEASE}} {{.OPENMETADATA_DEPS_CHART}} --namespace {{.OPENMETADATA_NAMESPACE}} --create-namespace --values {{.OPENMETADATA_DEPS_VALUES}} --wait --timeout 15m
+      - helm upgrade --install {{.OPENMETADATA_RELEASE}} {{.OPENMETADATA_CHART}} --namespace {{.OPENMETADATA_NAMESPACE}} --values {{.OPENMETADATA_VALUES}} --wait --timeout 15m
+      - kubectl -n {{.OPENMETADATA_NAMESPACE}} rollout status deploy/openmetadata --timeout=600s
+      - kubectl -n {{.OPENMETADATA_NAMESPACE}} get pods
+
+  openmetadata:undeploy:
```
libera13, line 210: As I see in `openmetadata:apply` you create also namespace and secret. What do you think about undeploying it here too?

plugins/cmd/openmetadata/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+FROM golang:1.26-alpine AS builder
+
+WORKDIR /app
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY plugins/pkg/ ./plugins/pkg/
+COPY plugins/internal/ ./plugins/internal/
+COPY plugins/cmd/openmetadata/ ./plugins/openmetadata/
```
libera13: please use ARG for pluginname, like in https://github.com/naira-project/naira/pull/83/changes#diff-d070a82c1a64713feea703e00f995335650161601056fcdb4bf28b55d4844f30

deploy/dev/infra/kubernetes/openmetadata-secrets.yaml
libera13, line 1: After adding namespace creation here, the filename `openmetedata-secrets` appears to be misleading

PR #95, opened 2026-07-13
feat: Keycloak integration into Naira

### LINKED ISSUE #24: Integrate Keycloak to Naira
In the first iteration, mock users can be used but after some time, this won't be enough as we would like to migrate the PoC to Naira project itself. So, for using OpenFGA's features, we also need to integrate Keycloak to our PoC to enhance a smooth migration.

kocakkaan:
## Description

This PR aims to integrate Keycloak with Naira. This PR is about https://github.com/naira-project/naira/issues/24

Upon discussion, it has been decided that Naira UI will be powered by OpenMFP portal which is under `naira-openmfp-portal`.

Further discussion is necessary for the following points:
* We may need to further think about how to handle more complex permissions - e.g. different granularity, asked by @akavel-reply .
* What approach are we taking for populating user access permissions?, asked by @libera13 and
* How to handle authorization coming from different plugins? , asked by @Daviidg 

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [x] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

## AI Usage

Claude Sonnet 4.6 used for refactoring and debugging.


catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
+package auth
+
+import (
+	"context"
+	_ "embed"
+	"encoding/json"
+	"fmt"
+
+	openfga "github.com/openfga/go-sdk"
+	. "github.com/openfga/go-sdk/client"
```
akavel-reply: It's not recommended to use dot-imports (it's ok to use import renames however if that can make things easier to understand later in the file)

[Effective Go - The Go Programming Language](https://go.dev/doc/effective_go#package-names)

> (Don't use the `import .` notation, which can simplify tests that must run outside the package they are testing, but should otherwise be avoided.)

and also:

[Go Wiki: Go Code Review Comments - The Go Programming Language](https://go.dev/wiki/CodeReviewComments#import-dot)

> Except for this one case, do not use import . in your programs. It makes the programs much harder to read because it is unclear whether a name like Quux is a top-level identifier in the current package or in an imported package.

catalog/internal/auth/keycloak.go
```diff
@@ -0,0 +1,105 @@
...
+
+// KeycloakConfig holds the client and realm needed for token verification.
+type KeycloakConfig struct {
+	Client TokenDecoder
+	Realm  string
+}
+
+// TokenClaims holds the user identity extracted from a verified Keycloak JWT.
+type TokenClaims struct {
+	Sub        string
```
akavel-reply: IIRC, I think you mentioned we could name this field `UserID` or something like that? I must admit `Sub` doesn't tell me anything, it's very unclear to me; ***however***, if that's a field that in fact has exactly such name in Keycloak JWT, then it could make sense to leave it as `Sub`; but in this case I'd be grateful if we could add some comment explaining in more detail what this field means.

It could make sense to have a call about this.

**EDIT:** From what I can see, `jwt.RegisteredClaims` struct uses field name `Subject` here - maybe that's a field name we could also use? Still, even then I'd be grateful for a comment, clarifying e.g. the difference between this and the `Username` field.

catalog/internal/auth/keycloak.go
```diff
@@ -0,0 +1,105 @@
...
+func NewAuthMiddleware(kc KeycloakConfig) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			authHeader := r.Header.Get("Authorization")
+			if authHeader == "" {
+				http.Error(w, "missing authorization header", http.StatusUnauthorized)
+				return
+			}
+
+			parts := strings.SplitN(authHeader, " ", 2)
```
akavel-reply: Apparently, Go now provides a helper `strings.Cut` function for cases like this: https://pkg.go.dev/strings#Cut

catalog/internal/auth/keycloak.go
```diff
@@ -0,0 +1,105 @@
...
+
+// TokenClaims holds the user identity extracted from a verified Keycloak JWT.
+type TokenClaims struct {
+	Sub        string
+	Email      string
+	Username   string
+	RealmRoles []string
+}
+
+type keycloakClaims struct {
```
akavel-reply: are we using this struct anywhere? I don't clearly see it used

catalog/internal/auth/keycloak/keycloak.go
```diff
@@ -0,0 +1,105 @@
...
+	tc, ok := ctx.Value(claimsKey).(TokenClaims)
+	return tc, ok
+}
+
+func parseTokenClaims(rawClaims *jwt.MapClaims) TokenClaims {
+	claims := *rawClaims
+	tc := TokenClaims{
+		Sub:      stringClaim(claims, "sub"),
+		Email:    stringClaim(claims, "email"),
+		Username: stringClaim(claims, "preferred_username"),
```
akavel-reply: Is there a reason to name this field `Username` rather than `PreferredUsername` (matching the claim name)? Those two names convey quite different meaning to me - which one is more approriate here semantically?

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+
```
akavel-reply: why separate the blocks of imports here? this is confusing when both blocks seem to have `github.com/openfga/...` entries

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+//go:embed model.fga.yaml
+var modelDSL string
+
+type Authorizer interface {
+	AuthorizeNodeRead(ctx context.Context, node catalog.NodeID, fgaModelType, fgaRelation string) error
+}
+
+type OpenfgaClient struct {
```
akavel-reply: Maybe: `OpenfgaClientWithModelID`?

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+	FgaClient  *OpenFgaClient
+	FgaModelID string
```
akavel-reply: The `Fga` name prefixes should not be necessary here - the struct already has `Openfga` in its name, so there's no need to repeat it.

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// AuthorizeNodeRead checks the existing OpenFGA tuples to determine whether the
+// authenticated caller has the given relation on the requested node.
+func (cli OpenfgaClient) AuthorizeNodeRead(ctx context.Context, node catalog.NodeID, fgaModelType, fgaRelation string) error {
```
akavel-reply: Again, no need to repeat `fga` here in argument names: we're in `OpenfgaClient` struct, so we already know we're generally in scope of fga topic; also, you mentioned "relation" in the godoc, so it is a good style to have an argument with matching name (see: https://go.dev/doc/comment#func).

On the other hand, `fgaModelType` (or `modelType`) is not clear to me here, and doesn't seem documented in the godoc: could you clarify its meaning, and how it should be used?

With that said, `fgaRelation` (or `relation`) is also still not super clear to me - if using this function, I would not be really sure what I should put there. Could you add some more explanation, maybe a link to OpenFGA docs, or whatever else is appropriate to clarify what specific values could/should be passed through that argument?

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+	claims, ok := ClaimsFromContext(ctx)
+	if !ok || claims.Sub == "" {
+		return fmt.Errorf("no authenticated user in request context")
+	}
+
+	fgaClient, err := cli.GetClient()
```
akavel-reply: `cli` name is confusing to me, as it makes me think of CLI (Command-Line Interface). It's quite common in Go to use one-letter names for the "receiver" (i.e. the type on which a method is called) within the method's definition. Let's please rename the receiver here to `c`:


```suggestion
	fgaClient, err := c.GetClient()
```

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return cli.FgaClient, nil
+}
+
+func (cli OpenfgaClient) GetModelID() (string, error) {
+	if cli.FgaClient == nil {
+		return "", fmt.Errorf("openfga client not initialized: call Setup first")
+	}
+	return cli.FgaModelID, nil
+}
```
akavel-reply: Can we drop those funcs, and just access `cli.FgaClient` / `cli.FgaModelID` where needed? It doesn't seem to me like we expect to have an `OpenfgaClient` with only one of the fields set in our code.

If we don't expect to set those fields manually, and we always require SetupOpenfgaClient() first, it'd be more typical to make those fields private.

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	if !allowed {
+		return fmt.Errorf("user %q is not allowed to %s %s", claims.Sub, fgaRelation, object)
+	}
+
+	return nil
+}
+
+func SetupOpenfgaClient(apiUrl, storeName string) (OpenfgaClient, error) {
```
akavel-reply: It would be more typical to name this function `NewOpenfgaClient()`, and also make it return a pointer to struct; and also, to put it right below or above the definition of the `OpenfgaClient` struct (it's a good style to try keeping functions "in chronological order" in a file, in the order in which they're expected to be used by the caller - this makes it easier to read and understand the code for a reviewer or someone who first enters the codebase).


```suggestion
func NewOpenfgaClient(apiUrl, storeName string) (*OpenfgaClient, error) {
```

catalog/internal/auth/openfga.go
akavel-reply, line 1: If possible, please move the types and functions related strictly to OpenFGA to a subpackage/subdirectory:

`catalog/internal/auth/openfga/`

Then, please remove occurrences of "Openfga" name from the types, functions, methods, and fields in this file:

- `OpenfgaClient` -> `Client`
- `SetupOpenfgaClient()` -> `NewClient()`
- `FgaModelID` -> `ModelID`

etc.

See: https://go.dev/doc/effective_go#package-names , https://go.dev/wiki/CodeReviewComments#package-names , etc.

catalog/internal/auth/keycloak/keycloak.go
akavel-reply, line 1: If possible, please move the types and functions related strictly to Keycloak to a subpackage/subdirectory:

`catalog/internal/auth/keycloak/`

Then, please remove occurrences of "Keycloak" name from the types, functions, methods, and fields in this file - for example:

- `KeycloakConfig` -> `Config`

etc.

See: https://go.dev/doc/effective_go#package-names , https://go.dev/wiki/CodeReviewComments#package-names , etc.

On the other hand, if some code is **NOT** strictly related to Keycloak, and can be useful more generically or for more clients, it could possibly make sense to keep it in `catalog/internal/auth/`. (But it's also fine to just start with an assumption that everything is specialized, and only start moving common code into a generic `catalog/internal/auth/` package later in the future, when we have more than one specialized implementation, and we really actually can then see which parts are generic, and which are not.)

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return OpenfgaClient{}, fmt.Errorf("Writing OpenFGA model: %w", err)
+	}
+
+	var cli OpenfgaClient
+
+	cli.FgaClient = client
+	cli.FgaModelID = modelId
+
+	return cli, nil
```
akavel-reply: It would be quite typical and idiomatic to write this code like so:


```suggestion
	return OpenfgaClient{
		FgaClient:  client,
		FgaModelID: modelId,
	}, nil
```

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+	storeId, err := createStore(client, storeName)
+	if err != nil {
+		return OpenfgaClient{}, fmt.Errorf("creating a store with the client: %w", err)
+	}
+
+	if err := client.SetStoreId(storeId); err != nil {
+		return OpenfgaClient{}, fmt.Errorf("setting openfga store id: %w", err)
+	}
+
+	modelId, err := writeOpenFGAModel(client, jsonStr)
```
akavel-reply: Do we really want to do it every time catalog is restarted? is it expected to work? is it a good/recommended practice to do it like so?

Is there some recommended approach to versioning and migrating/upgrading OpenFGA models? if there is, we don't need to already implement it at our early stage, but if there is, it'd be good to at least mention it in a TODO comment somewhere in this file.

catalog/internal/auth/openfga.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return OpenfgaClient{}, fmt.Errorf("Writing OpenFGA model: %w", err)
```
akavel-reply: Please try to keep those messages starting from small letter - same also in a few other places in this file


```suggestion
		return OpenfgaClient{}, fmt.Errorf("writing OpenFGA model: %w", err)
```

catalog/internal/auth/keycloak.go
kocakkaan: Deleted.

catalog/internal/auth/keycloak/keycloak.go
kocakkaan: `preferred_username` is the field we are getting from Keycloak when we parse token claims.

catalog/internal/auth/keycloak.go
kocakkaan: Changed the field here to `UserID`.

catalog/internal/auth/openfga.go
kocakkaan: About the second paragraph, I don't have an idea about it but it would be an interesting research. I am having a look and also will provide a TODO comment in the file. About the first paragraph, now if there is already a store it returns the existing store. I am also implementing the second part where we also need to see if there is already an existing model so we won't have more one contextually same model.

catalog/internal/auth/openfga.go
kocakkaan: `sdk "github.com/openfga/go-sdk"` and `openfga "github.com/openfga/go-sdk/client"` are the current imports, I have deleted the `import .` notation.

libera13:
Few parts from PR description:
"will be using the token retrieved from Platform Mesh's organizations"
I am not exactly sure what this means. Do you mean that we will use their Keycloak?

"We have a separate OpenFGA instance running inside the same cluster."
I think it is worth considering two paths. 
- in Platform Mesh mode, we should use their services (i.e., their OpenFGA and Keycloak; in that case, we would need to understand how to leverage their OpenFGA). 
- in standalone mode, let it be Keycloak alone (without OpenFGA), or Keycloak combined with OpenFGA (or another authorization system)

--- 

Why are we using ReBAC if we are implementing a RBAC? If we see a specific case where we actually need granular permissions—for example, a flow where this can be automatically populated—please mention it. Otherwise, I feel we could just use RBAC.

Returning to the big picture: what do we want to authorize? What will roles likely look like in Naira? As I see for example we can have an admin who can launch/add plugins, and users who can see almost everything but only have edit rights for a minimal set of things. Maybe here would be few groups of users depending what they can see. When we will have more actions in Naira then I think it will be good to rethink our approach.

RBAC should be enough here. Otherwise, why complicate things with ReBAC?

OpenFGA makes sense when there are relationships between objects. That is how I understand why they used OpenFGA in Platform Mesh. For example, if there are KCP workspaces:

- Workspace A
  - Workspace AA
    - db-postgres
    - db-mysql
  - Workspace AB
      - db-mongo
- Workspace B
    - db-cassandra

If I want to ensure that if a user has 'viewer' access to Workspace A, they also have 'viewer' access to all data under it (e.g. db-mysql), then this can be handled with a single tuple: "User-Adam viewer Workspace A." It is clean and simple. ReBAC makes sense in a world where there is a clear hierarchy.

I propose to remove the OpenFGA part from the PR for now. I think that before utilizing OpenFGA, we should first find a good use case for it. For example, it would maybe make sense to use platform-mesh's OpenFGA. The exact scenario I envision is that the 'platform-mesh plugin' pulls data from it—e.g., some CRs—and Naira knows how to talk to platform-mesh's OpenFGA in a way that it doesn't even add its own tuples there, but can just read. In the same way graphql-gateway checks if a user has access. Only what would be needed to map Naira nodes to the GraphQL-gateway's way of checking data

catalog/internal/auth/keycloak.go
rootlsz: there was a comment on a chat pointing to this resource:
- https://www.keycloak.org/docs/latest/server_development/index.html#_action_token_anatomy 
> sub - ID of the user

catalog/internal/auth/keycloak/authz.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/naira-project/naira/plugins/pkg/pluginapi"
+)
+
+const (
+	RealmRoleAIEngineer          = "ai-engineer"
+	RealmRoleApplicationEngineer = "application-engineer"
+)
+
+var realmRoleByNodeKind = map[string]string{
+	pluginapi.NodeKindModel:   RealmRoleAIEngineer,
```
akavel-reply: so, application engineer can't see models? this is surprising to me, I thought we wanted a dev to be able to see models in Naira?

catalog/internal/auth/keycloak/authz.go
```diff
@@ -0,0 +1,43 @@
...
+	pluginapi.NodeKindDataset: RealmRoleAIEngineer,
+}
```
akavel-reply: what about other node kinds we already have? and what will happen when new plugins introduce new node kinds?

catalog/internal/auth/keycloak/authz.go
```diff
@@ -0,0 +1,43 @@
...
+
+func requiredRealmRoleForNodeKind(kind string) string {
+	if role, ok := realmRoleByNodeKind[kind]; ok {
+		return role
+	}
+	return RealmRoleApplicationEngineer
```
akavel-reply: why? this is confusing and surprising to me; what's the core idea behind the authorization scheme this PR is currently trying to propose? is it described somewhere?

catalog/internal/auth/keycloak/keycloak.go
```diff
@@ -0,0 +1,112 @@
...
+
+// TokenClaims holds the user identity extracted from a verified Keycloak JWT.
+type TokenClaims struct {
+	UserID     string
+	Email      string
+	Username   string
+	RealmRoles []string
+}
+
+func NewAuthMiddleware(kc Config) func(http.Handler) http.Handler {
```
akavel-reply: Please rename `kc` to `cfg`, will be more clear (`kc` keeps confusing me in body of this func)


```suggestion
func NewAuthMiddleware(cfg Config) func(http.Handler) http.Handler {
```

catalog/internal/auth/keycloak/keycloak.go
```diff
@@ -0,0 +1,112 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if kc.Client == nil {
```
akavel-reply: I think we should do this check in NewAuthMiddleware func, not in the returned func - this looks like a misconfiguration, I would expect we want to know about it early when starting this app, rather than only much later when the first request comes

catalog/internal/auth/keycloak/keycloak.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+			tokenString := strings.TrimSpace(after)
+
+			_, rawClaims, err := kc.Client.DecodeAccessToken(r.Context(), tokenString, kc.Realm)
+			if err != nil || rawClaims == nil {
+				writeJSONError(w, http.StatusUnauthorized, "invalid token")
+				return
+			}
+
+			tc := parseTokenClaims(rawClaims)
```
akavel-reply: please rename `tc` to `claims`


```suggestion
			claims := parseTokenClaims(rawClaims)
```

catalog/internal/auth/keycloak/keycloak.go
akavel-reply: Please add godoc comment explaining what this function does - the name alone might be misleading (for example, I thought maybe it checks permissions to a page somehow - but it only loads them and stores into context)

catalog/internal/auth/keycloak/keycloak.go
akavel-reply: Sure - but my question is different, why our field name `Username` differs from the name we extract from Keycloak - I'd expect `PreferredUsername` field here:


```suggestion
		PreferredUsername: stringClaim(claims, "preferred_username"),
```

catalog/internal/auth/keycloak/authz.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// AuthorizeNodeRead checks that the authenticated caller's Keycloak realm
+// roles grant them access to nodes of the given kind.
+func AuthorizeNodeRead(ctx context.Context, kind string) error {
+	claims, ok := claimsFromContext(ctx)
+	if !ok || claims.UserID == "" {
```
akavel-reply: What is the reason for this check of UserID? Is it not enough to just check RealmRoles as we do below?

catalog/internal/httpapi/router_test.go
```diff
@@ -236,3 +276,93 @@ func TestRunAllPluginsReturnsPluginErrorsInResults(t *testing.T) {
...
+		expectedStatusCode int
+	}{
+		{
+			name:               "ai engineer can read a model node",
+			bearerToken:        aiEngineerBearerToken,
+			path:               "/v1/nodes/model/mlflow/fraud-detector",
+			expectedStatusCode: http.StatusOK,
+		},
+		{
+			name:               "ai engineer cannot read an application node",
```
akavel-reply: that is strange to me - I thought we want AI engineers to see what apps are using their models, am I misunderstanding something?

go.mod
```diff
@@ -3,7 +3,9 @@ module github.com/naira-project/naira
 go 1.26.0
 
 require (
-	github.com/go-chi/chi/v5 v5.3.0
+	github.com/Nerzal/gocloak/v13 v13.9.0
+	github.com/go-chi/chi/v5 v5.2.3
```
akavel-reply: why downgrade go-chi version?

ui-poc/Dockerfile
```diff
@@ -10,7 +10,7 @@ FROM nginx:1.30-alpine
 COPY --from=builder /app/build /usr/share/nginx/html
 COPY nginx.conf.template /etc/nginx/templates/default.conf.template
 COPY docker-entrypoint.sh /custom-entrypoint.sh
-RUN chmod +x /custom-entrypoint.sh
+RUN sed -i 's/\r//' /custom-entrypoint.sh && chmod +x /custom-entrypoint.sh
```
akavel-reply: what's the point of this change? I'm surprised and confused

kocakkaan: This was because I am working with WSL and I always had encountered "line endings" problem. Instead of doing it every time when running Docker manually, I put the script here.

akavel-reply: Is it possible you could adjust it in your local git settings instead? https://stackoverflow.com/a/4182346
and/or in your local editor settings (so that the editor doesn't write CRs)?
I'm not aware of such a sed command being a common pattern in Dockerfiles; I believe the industry standard is to assume LFs in files, and modern tools tend to have ways/options that should make sure LFs with no CRs are used when writing files.

akavel-reply:
**NOTE: To anyone who might take over this PR:**

We discussed further with Kaan and decided, that this PR should be dialed back, so that for now we will only use Keycloak for authentication, *not* for authorization. So, we only want a check if a user is logged in - then the API endpoints should be available to them (with all data); or, not logged in - then the API should not be available.

ui-poc/Dockerfile
kocakkaan: Removed. 👍

go.mod
kocakkaan: Upgraded again. 👍

rootlsz (approved):
👏

catalog/internal/auth/keycloak/keycloak.go
```diff
@@ -0,0 +1,112 @@
...
+func writeJSONError(w http.ResponseWriter, status int, message string) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	_ = json.NewEncoder(w).Encode(map[string]string{"error": message})
+}
+
+func claimsFromContext(ctx context.Context) (TokenClaims, bool) {
+	claims, ok := ctx.Value(claimsKey).(TokenClaims)
+	return claims, ok
+}
```
Daviidg: Is this needed? Doesnt get called anywhere?

Daviidg (approved):
Code looks good. I like the idea of moving auth checks to a middleware.

catalog/internal/auth/keycloak/keycloak.go
kocakkaan: It was from previous iteration, deleted it. Thank you!

PR #125, opened 2026-08-03
feat: support configurable node path prefix per plugin instance

rootlsz:
## Description

Removed the `pluginName` constant and replaced it with a configurable `PATH_PREFIX` environment variable.

image

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [ ] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

### DIFF
--- a/deploy/dev/infra/kubernetes/catalog.yaml
+++ b/deploy/dev/infra/kubernetes/catalog.yaml
@@ -189,8 +189,6 @@ spec:
               value: "50053"
             - name: DEPL_CALLS_SVC_KUBECONFIG
               value: ""
-            - name: PATH_PREFIX
-              value: depl-calls-svc
           resources:
             requests:
               cpu: 50m
@@ -209,8 +207,6 @@ spec:
               value: ""
             - name: DEPL_USES_LITELLM_NAMED_HOSTS
               value: "litellm=http://litellm.litellm:4000"
-            - name: PATH_PREFIX
-              value: depl-uses-litellm
           resources:
             requests:
               cpu: 50m
@@ -225,8 +221,6 @@ spec:
           env:
             - name: PORT
               value: "50055"
-            - name: PATH_PREFIX
-              value: fluxcd
           resources:
             requests:
               cpu: 50m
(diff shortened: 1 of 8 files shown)


deploy/dev/infra/kubernetes/catalog.yaml
```diff
@@ -153,6 +153,8 @@ spec:
                 secretKeyRef:
                   name: catalog-secrets
                   key: LITELLM_API_KEY
+            - name: PATH_PREFIX
+              value: litellm-path-prefix
```
libera13: To have more clean UI
```suggestion
              value: litellm
```
in others too

plugins/cmd/depl_calls_svc/main.go
```diff
@@ -25,14 +25,13 @@ import (
...
 
-const pluginName = "depl_calls_svc"
-
 var (
 	gvrDeployments = schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
 	gvrServices    = schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}
 )
 
 type config struct {
+	PathPrefix string `env:"PATH_PREFIX,required"`
```
libera13: here this value is not used as PathPrefix but as pluginName for logging, so looks like not needed change. I guess the same with other depl_uses... and fluxcd plugin

plugins/cmd/litellm/main.go
```diff
@@ -29,6 +27,7 @@ const (
 )
 
 type config struct {
+	PathPrefix  string        `env:"PATH_PREFIX,required"`
```
libera13: I think it should have default value in all plugins you changed

rootlsz: we discussed it before and you said that in k8s it should fail if someone doesn't set this, and in tests we'll set the required env variable. Why do we want to change this now?

akavel-reply (approved):
LGTM

plugins/cmd/litellm/main.go
libera13: At the time, I was thinking about the “hostname–URL” pair. Here, only the prefix is specified. It seemed to me that the default URL didn't make much sense.

I think the prefix could have a default value, which would make running it locally easier

rootlsz: added default value

libera13 (approved):

Issue #9, opened 2026-05-05
Reference Integration: AI API Gateway (LiteLLM)


mkorbi:
## 📋 Overview

**User Story:** As an AI Engineer I want to gain fast an overview which Models, MCPs and skills are already available.

---

## 🎯 Context
A core outcome of Naira is the integration with an AI API Gateway to ensure secure, observable, and compliant consumption of AI services. Naira connects to LiteLLM as a reference AI gateway integration, synchronizes skills and provider metadata, and shows gateway context in the AI Engineering Hub. LiteLLM remains the runtime gateway.
To validate the Plugin Framework built in Milestone 2, we need reference integrations. LiteLLM is the currently proposed reference implementation for the gateway. This feature focuses on wrapping LiteLLM as a Naira plugin, demonstrating how the platform orchestrates gateway configuration dynamically.

---

### 1. Plugin Integration
The plugin publishes discovered models, MCPs and skills back to Naira, ensuring that the AI Gateway endpoints are visible in the platform's unified catalog.

### 2. Models
Listen the models in the model catalog, enrich it with available information, and identify if those models are already somehow else listed e.g. by an OpenAI API.

### 3. MCPs
Listen the available MCPs in an own new catalog. Indicate where those come from and on a detailed page, show which tools does it provide. This is related to https://github.com/naira-project/naira/issues/26

---

## 🛠 Functional Requirements

- **Plugin Installation:** The LiteLLM plugin can be installed via the Naira Plugin Lifecycle Manager.
- **Provider Configuration:** Users can configure LLM provider credentials (e.g., OpenAI, Azure, Anthropic), which the plugin securely passes to LiteLLM, and becomes visible at Naira.
- **Catalog Sync:** The plugin registers configured endpoints.

---

## 🛡️ Non-Functional Requirements (NFRs)

- **Performance:** The configuration sync loop from Naira to the LiteLLM instance must apply changes within seconds without dropping active connections.
- **Security:** Provider credentials must be handled securely via the Secret Management framework.
- **Extensibility:** The plugin structure must be generic enough to serve as a blueprint for integrating alternative gateways in the future.

---

## ✅ Acceptance Criteria
- [x] The LiteLLM plugin is successfully installed via the Plugin Framework.
- [x] in LiteLLM configured Models appear in the model catalog
- [x] in LiteLLM configured MCPs appear in the model catalog
- [ ] A skill is configured and visible via Naira platform.

libera13:
I have few questions

[...]

PR #139, opened 2026-08-10
feat: OpenAI api plugin

### LINKED ISSUE #119: [post-hackathon] Extract shared "OpenAI-API-models" plugin from Thalamus PR
During the hackathon, based on the submitted PR #117 , we noticed that both it, and our preexisting litellm plugin, fetch data from the standard OpenAI API `/v1/models` endpoint.

We should move this common code to a generic "openai-api-models" plugin.

- If there's no other code left in the litellm plugin afterwards, we should remove the litellm plugin.
- We should then explain in this new plugin's readme that it's: "compatible with: LiteLLM, Thalamus, others".
- The new plugin should not use "litellm/" prefix in path; instead, it should allow the user to specify a custom prefix, so that they could e.g. use "litellm/" for litellm, and "thalamus/" for Thalamus, or other. See e.g. the `DEPL_USES_LITELLM_NAMED_HOSTS` env var in the `depl_uses_litellm` plugin.
- We might also need to similarly adjust the `depl_uses_litellm` plugin name and code to `depl_uses_oai_models` or something like that (draft name; ideally, the name should be consistent with the base plugin, and yet both should still be understandable).

After the refactoring described above, the Thalamus plugin should still probably keep and use the code related to OpenAI API: it adds its own CRD-related spin on that, which is the important part, but we probably make it easy to use just the Thalamus plugin with Thalamus as a shrinkwrapped package, rather than requiring users to manually set-up also a generic OpenAI API plugin for the same Thalamus instance. (Though, I'm still unsure on that; we do still separate the "thalamus" vs. "deployment_uses_thalamus" plugins, so there's already precedent for multiple plugins... so maybe let's only keep the CRD stuff in the Thalamus plugin, and document in the readme that it should be used together with the generic openai-api-models plugin?)

Daviidg:
## Description

Code introduces OpenAI api compatible plugin. For now only `v1/models` is covered. Code has been tested and works for littellm, vllm and llamacpp. Original ticket is https://github.com/orgs/naira-project/projects/2/views/8?sliceBy%5Bvalue%5D=Daviidg&pane=issue&itemId=219749480&issue=naira-project%7Cnaira%7C119, which originated from the Thalamus openai spec implementation

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [x] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

Describe how and for what AI was used in this PR
## AI Usage
Used Sonnet 5 for code + some automated testing

Screenshot 2026-08-14 at 10 15 13

### DIFF
--- a/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
+++ b/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
@@ -185,42 +185,6 @@ spec:
             limits:
               cpu: 200m
               memory: 128Mi
-        - name: plugin-openai-api-models-vllm
-          image: openai-api-models:local
-          restartPolicy: Always
-          imagePullPolicy: IfNotPresent
-          env:
-            - name: PORT
-              value: "50058"
-            - name: OPENAI_API_MODELS_BASE_URL
-              value: http://vllm-opt125m.vllm.svc.cluster.local:8001
-            - name: PATH_PREFIX
-              value: openai_vllm
-          resources:
-            requests:
-              cpu: 50m
-              memory: 64Mi
-            limits:
-              cpu: 200m
-              memory: 128Mi
-        - name: plugin-openai-api-models-llamacpp
-          image: openai-api-models:local
-          restartPolicy: Always
-          imagePullPolicy: IfNotPresent
-          env:
-            - name: PORT
-              value: "50059"
-            - name: OPENAI_API_MODELS_BASE_URL
-              value: http://llama-qwen25-05b.llamacpp.svc.cluster.local:8080
-            - name: PATH_PREFIX
-              value: openai_llamacpp
-          resources:
-            requests:
-              cpu: 50m
-              memory: 64Mi
-            limits:
-              cpu: 200m
-              memory: 128Mi
         - name: plugin-mlflow
           image: mlflow:local
           restartPolicy: Always
@@ -320,7 +284,7 @@ spec:
             - name: PORT
               value: "8090"
             - name: PLUGIN_ADDRESSES
-              value: "litellm=localhost:50051,mlflow=localhost:50052,depl-calls-svc=localhost:50053,depl-uses-litellm=localhost:50054,fluxcd=localhost:50055,openmetadata=localhost:50056,openai-api-models-litellm=localhost:50057,openai-api-models-vllm=localhost:50058,openai-api-models-llamacpp=localhost:50059"
+              value: "litellm=localhost:50051,mlflow=localhost:50052,depl-calls-svc=localhost:50053,depl-uses-litellm=localhost:50054,fluxcd=localhost:50055,openmetadata=localhost:50056,openai-api-models-litellm=localhost:50057"
           ports:
             - containerPort: 8090
               name: http
--- a/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
+++ b/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
@@ -171,7 +171,7 @@ spec:
               value: "50057"
             - name: OPENAI_API_MODELS_BASE_URL
               value: http://litellm.litellm.svc.cluster.local:4000
-            - name: OPENAI_API_MODELS_NODE_PREFIX
+            - name: PATH_PREFIX
               value: openai_litellm
             - name: OPENAI_API_MODELS_API_KEY
               valueFrom:
--- a/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
+++ b/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
@@ -162,6 +162,29 @@ spec:
             limits:
               cpu: 200m
               memory: 128Mi
+        - name: plugin-openai-api-models
+          image: openai-api-models:local
+          restartPolicy: Always
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: PORT
+              value: "50057"
+            - name: OPENAI_API_MODELS_BASE_URL
+              value: http://litellm.litellm.svc.cluster.local:4000
+            - name: OPENAI_API_MODELS_NODE_PREFIX
+              value: openai_litellm
+            - name: OPENAI_API_MODELS_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: catalog-secrets
+                  key: LITELLM_API_KEY
+          resources:
+            requests:
+              cpu: 50m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
         - name: plugin-mlflow
           image: mlflow:local
           restartPolicy: Always
@@ -261,7 +284,7 @@ spec:
             - name: PORT
               value: "8090"
             - name: PLUGIN_ADDRESSES
-              value: "litellm=localhost:50051,mlflow=localhost:50052,depl-calls-svc=localhost:50053,depl-uses-litellm=localhost:50054,fluxcd=localhost:50055,openmetadata=localhost:50056"
+              value: "litellm=localhost:50051,mlflow=localhost:50052,depl-calls-svc=localhost:50053,depl-uses-litellm=localhost:50054,fluxcd=localhost:50055,openmetadata=localhost:50056,openai-api-models=localhost:50057"
           ports:
             - containerPort: 8090
               name: http
--- a/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
+++ b/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
@@ -162,7 +162,7 @@ spec:
             limits:
               cpu: 200m
               memory: 128Mi
-        - name: plugin-openai-api-models
+        - name: plugin-openai-api-models-litellm
           image: openai-api-models:local
           restartPolicy: Always
           imagePullPolicy: IfNotPresent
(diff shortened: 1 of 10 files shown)


deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
```diff
@@ -162,6 +162,65 @@ spec:
...
+              memory: 128Mi
+        - name: plugin-openai-api-models-vllm
+          image: openai-api-models:local
+          restartPolicy: Always
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: PORT
+              value: "50058"
+            - name: OPENAI_API_MODELS_BASE_URL
+              value: http://vllm-opt125m.vllm.svc.cluster.local:8001
```
libera13: From one side I understand it proves the point that plugin is generic. But in this setup does it make sense to ask vllm for models? Giving exact url looks like manual job for each vllm/llama.cpp instance

deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
```diff
@@ -261,7 +320,7 @@ spec:
             - name: PORT
               value: "8090"
             - name: PLUGIN_ADDRESSES
-              value: "litellm=localhost:50051,mlflow=localhost:50052,depl-calls-svc=localhost:50053,depl-uses-litellm=localhost:50054,fluxcd=localhost:50055,openmetadata=localhost:50056"
+              value: "litellm=localhost:50051,mlflow=localhost:50052,depl-calls-svc=localhost:50053,depl-uses-litellm=localhost:50054,fluxcd=localhost:50055,openmetadata=localhost:50056,openai-api-models-litellm=localhost:50057,openai-api-models-vllm=localhost:50058,openai-api-models-llamacpp=localhost:50059"
```
libera13: so many new plugins, makes me think that we will need something else then this side containers

plugins/cmd/openai_api_models/main.go
```diff
@@ -0,0 +1,120 @@
...
+	logger     *log.Logger
+	config     config
+	nodePrefix string
+}
+
+func New(config config, logger *log.Logger) (*Plugin, error) {
+	if strings.TrimSpace(config.BaseURL) == "" {
+		return nil, fmt.Errorf("no endpoint configured: OPENAI_API_MODELS_BASE_URL is empty")
+	}
+	// The prefix is what keeps models of different endpoints apart
```
libera13: is this comment needed? Doesn't say too much to me

plugins/cmd/openai_api_models/main.go
libera13, line 1: Can we remove litellm plugin?

plugins/cmd/openai_api_models/main.go
```diff
@@ -0,0 +1,120 @@
...
+	prefix := strings.Trim(strings.TrimSpace(config.PathPrefix), "/")
```
libera13: Does this normalization make sense? In my opinion, it somewhat masks the problem. I believe that if someone enters the `litellm/` prefix, they should receive an error message. I would simply remove this normalization here. And I would expect users to be more aware of the configurations they’re entering.

plugins/cmd/openai_api_models/main_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	p, err := New(config{BaseURL: mockServer.URL, PathPrefix: "openai_llamacpp"}, testLogger())
+	require.NoError(t, err)
+	p.httpClient = mockServer.Client()
+
+	res, err := p.Collect(context.Background())
+	require.NoError(t, err)
+
+	// Ids that collapse to nothing are skipped, and no path gains a second segment
+	// - the UI reads the second-to-last segment as the source.
```
libera13: UI can change. I think plugins should not care too much about it. And this comment can be quickly misleading

plugins/internal/openaiutil/models.go
```diff
@@ -0,0 +1,69 @@
+// Package openaiutil provides a small client for the parts of the OpenAI API
+// that Naira plugins need, so that every plugin talking to an OpenAI-compatible
+// endpoint shares one implementation.
```
libera13, line 3: for now only one plugin use it? If so then for me it still fits more to cmd/openai_api_models dir

libera13 (commented):
Let's say I skimmed through the PR :)

plugins/internal/openaiutil/models.go
Daviidg: I spoke with Mateusz about it. I wanted to do more of a refactoring and move some of the openai logic into a utils file that plugins like littellm, openai, etc could use. However, he suggested just to just implement the openai plugin for now and refactor later if needed

plugins/cmd/openai_api_models/main.go
Daviidg: At the moment yes. Litellm follows openai spec without adding extra fields for models. However I would suggest keeping it. I am also working on the integration of LitteLLM Mcp (https://github.com/orgs/naira-project/projects/2/views/8?sliceBy%5Bvalue%5D=Daviidg&pane=issue&itemId=5150387592&issue=naira-project%7Cnaira%7C150) and that part isnt covered by openai spec, so we still need to keep a littellm specific plugin

deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
Daviidg: I essentially its meant to test it works with different providers. Same issue would happen if you made a vllm or llamacpp specific plugins I think. Imo the solution to reduce manual job needs to come from some sort of discovery mechanism, which was not the focus of this PR

deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
Daviidg: I also agree here

plugins/cmd/openai_api_models/main.go
libera13: Will you keep the fetching models in litellm? If yes maybe refactor it so it's uses your extracted utils to handle openai models

plugins/internal/openaiutil/models.go
libera13: I will suggest to refactor litellm plugin now :) otherwise we will have duplicated code

deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
libera13: I just don't see a reason why to have models listed from vllm/llama.cpp. As I understand it it will be always one model listed per plugin. 
I don't say openai generic plugin is not needed and that I want to create vllm/llamacpp specific plugins. No.

Maybe in some way the solution is to pass list of urls. So we have one plugin scraping data from multiple sources at once. 
Or maybe in this local dev cluster, don't call this vllama/llama.cpp services...

I just see that this approach proves the point that it works, which is cool. But having this in main will suggest that it's our way to do the things. And I think if we would like to really get models from vllm/llamacpp it should be more automatic, so it's not require manually specified URLs (I think they change often)

Probably the use case here is also, that normally this vllm models are registered in company ai gateway, so we would get them from e.g. litellm anyway

plugins/cmd/openai_api_models/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# --- Stage 2: Final ---
+FROM gcr.io/distroless/static-debian12:latest
+
+COPY --from=builder /out/plugin /plugin
+
+EXPOSE 50051
+
+USER nonroot:nonroot
+
+ENTRYPOINT ["/plugin"]
```
kocakkaan, line 26: I have seen that we use this template across all plugins in the identically. Do you think that we should make this a shared template across plugins, with PLUGIN argument being the sole variable that is added in Taskfiles? @Daviidg @libera13

plugins/cmd/openai_api_models/main.go
Daviidg: Refactored part of the code of litellm to openai util file. However I could not refactor too much code. The part of littellm that is openai compatible is the v1/models which we use only to list the models. With that info we then call the key/info request that extract littellm specific and not openai compatible properties

deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
Daviidg: Remomed the openai sidecars with llamacpp and vllm. 

I agree with you on the last point mentioning that ideally companies would have some model catalog like litellm we would integrate instead of vllm or llamacpp directly, unfortunatelly I can tell you that sometimes is not the case. Therefore I think it still is usefull to have an openai plugin as an option, even if it just connects to one model

plugins/cmd/litellm/main.go
```diff
@@ -58,7 +59,8 @@ func main() {
 }
 
 func (p *Plugin) Collect(ctx context.Context) (pluginapi.CollectResponse, error) {
```
libera13, line 61: I would remove completely the Litellm plugin. This AppIdentities CRD concept, was PoC. Plugin depl_uses_litellm made this idea obsolete.
And then maybe we can move code from openaiutil to plugin package

kocakkaan: I would say maybe not now but then after we evaluate litellm's position in [this ticket](https://github.com/naira-project/naira/issues/9). What do you think? @libera13

libera13: Maybe then we can get it back...? I'm just in favor of keeping it clean. And as litellm plugin doesn't have too much sense now, I propose to remove it

kocakkaan: You are right 👍 . We can remove it for now then.

plugins/cmd/openai_api_models/Dockerfile
kocakkaan: Irrelevant for now, therefore resolving the conversation.

Daviidg:
Quick summary on the openai plugin work in this PR:
- Implementation of openai utils functions that all plugins can use
- OpenAI API plugin that works with any OpenAI compatible server (vllm, llamacpp).
- Part of littellm plugin refactored to use the openai functions, however code can not be completely refactored because they use `/key/info` to get littellm specific info about the models

Limitations:
-  For now, only `v1/models` endpoint of OpenAI spec are implemented. This essentially lists the available models. However the OpenAI api spec has a lot more useful endpoints like `/v1/fine_tuning/jobs`, `/v1/batches`, `/v1/vector_stores`, `/v1/files` that could be implemented as these concepts get added to Naira

Issue #103, opened 2026-07-14
Improve generic table functionality


libera13:
Building tables yourself is complex. Consider using open-source solutions instead. Check if shadcn/ui has a table component, or explore other options. The table should support at least: sorting, filtering, changing column visibility, and changing column width.

PR #152, opened 2026-08-14
feat(ui): kind based page separation

### LINKED ISSUE #142: Create pages for implemented assets
Right now, the OpenMFP portal consists of one page called *Dashboard*. However, it is necessary to divide the assets into separate pages so that user would not feel overwhelmed about the actions. 

Solution is to keep the "Dashboard" page as an aggregated space where the user can navigate themselves into specific assets. Then, the user can navigate themselves to separate pages through OpenMFP portal.

kocakkaan:
## Description

This PR aims at providing asset based page separation of plugins and kinds. There are two implemented asset pages for now: `Model Catalog`, `Software Catalog` and `Dataset Catalog`. Each page consists of a subset of plugins which are connected with these assets. The empty state is also implemented, with the link to necessary plugins to be run in order to fill the specific catalog page with data. 

Related issue is #142.

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- [x] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

## AI Usage

Claude Sonnet 5 used 

### Overview page
image

### Model catalog
image

### Software catalog
image

### Dataset catalog (empty state as an example)
image

### Separate plugins page
image


mkorbi:
So it not only seperates finaly the catalogs but also shows which plugin per catalog is relevant?

kocakkaan:
> So it not only seperates finaly the catalogs but also shows which plugin per catalog is relevant?

The mapping of plugins and catalogues related to them are done manually in `ui-poc/src/config/viewpoints.ts`. I could not find a way to do it automatically, but in the future this can be done via an AI agent maybe. It can at least tell the user which plugin could be related to which catalog. @mkorbi

ui-poc/src/hooks/usePluginOperations.ts
```diff
@@ -200,6 +201,40 @@ export function usePluginsStatus(): UsePluginsStatusResult {
     }
   }, [settleOne, addError, token]);
 
+  /**
+   * Like `runAll`, but scoped to a specific set of plugins — used by viewpoints
+   * (e.g. Model, Software Catalog) that only want to trigger the plugins they show,
+   * rather than every plugin the backend knows about.
+   */
+  const runSubset = useCallback(
```
rootlsz, line 207: Could we avoid waiting for all plugins to finish before processing their results?

Right now, we have two steps:
1. `Promise.allSettled()` waits for all `apiRunPlugin()` calls to finish
2. then `Promise.all()` starts `settleOne()` for the results

Do we really need these two separate blocks, or could we process each result as soon as its `apiRunPlugin()` finishes, using one parallel pipeline?

For example:

`apiRunPlugin() → settleOne()`

for each plugin independently, and then wait for all pipelines to finish.

If there is a reason for keeping current implementation (I assume there is :D) could you explain why? 👀

mkorbi: hmm couldn't the plugins periodically just call updates, and the sync button is more like "I want to see a change that happened right now"?

libera13: Are you referring to the addition of a scheduling feature? In other words, if the plugin's schedule should be configured? For example, can a user specify whether the plugins should run once a day, once an hour, etc.?

kocakkaan: @rootlsz I was actually trying to mimic the behavior in `runAll` method on top of it, but I oversaw this problem. You are right, this would only force a sequential flow. I am changing this for `apiRunPlugin() → settleOne()` per-plugin.

mkorbi: @libera13 good point, true, please link it and forget what I said :D

ui-poc/src/components/PluginsManagerDialog.tsx
```diff
@@ -10,6 +10,8 @@ interface PluginsManagerDialogProps {
   open: boolean;
   onClose: () => void;
   onRunsCompleted: () => void;
+  /** Restrict the dialog to this subset of plugins (e.g. the plugins relevant to the current viewpoint). Omit to show every plugin. */
+  allowedPlugins?: string[];
```
libera13: I'm not convinced if plugins filtering in a dialog makes sense to me.

In the end, I think this plugin manager component can land on its own page. I would rather suggest just showing plugin names on empty states.

For example, on the datasets page, when a user doesn't have any datasets in the catalog yet, we can just write in the empty state: "We have an OpenMetadata plugin which can pull datasets for you," etc.

ui-poc/src/components/PluginTabs.tsx
```diff
@@ -0,0 +1,44 @@
...
+  activePlugin: string | null;
+  onSelect: (plugin: string | null) => void;
+}
+
+/**
+ * Underline-style tab bar shown below the kind selector: an "All" tab plus one
+ * tab per plugin that has claimed at least one node of the active kind.
+ * Selecting a plugin filters the table to nodes claimed by that plugin.
+ */
+export default function PluginTabs({ plugins, activePlugin, onSelect }: PluginTabsProps) {
```
libera13, line 14: I like the value which these tabs add; they are useful. But as I imagine we would have more advanced filtering, sorting, and pagination on the table, we can maybe think more generally about it

ui-poc/src/App.tsx
```diff
@@ -10,6 +11,21 @@ function App() {
         <Routes>
           <Route path="/" element={<CatalogView />} />
           <Route path="/catalog" element={<CatalogView />} />
+          {CATALOG_VIEWPOINTS.map((viewpoint) => (
```
libera13, line 20: I wonder if this generic catalogview is enough. Or rather we should go into direction of creating dedicated pages for "software catalog", "models" etc. As I see on Max's designs, the one shared element will be the table and filters. 

Maybe we just should move out this "Plugins & Ingestion" from CatalogView to separate page...

libera13 (commented):
The PR is relatively small, but it touches a few areas which I think we need to discuss a little bit :)

ui-poc/src/App.tsx
kocakkaan: you are right, this `Plugins & Ingestion` part would rather work as a separate page than embedded inside `CatalogView` and then we can have dedicated content changing with each asset. 👍

rootlsz (approved):
🔥

ui-poc/src/components/PluginTabs.tsx
kocakkaan: To be discussed further with the issue #103.

ui-poc/src/App.tsx
Daviidg: I think we should definitely go into dedicated pages for each tab. I am working for example on the MCP catalog, and there I would need already a way to display the tools each mcp has which dont really match well with the current generic implementation

Daviidg (approved):
Code looks good

ui-poc/src/hooks/usePluginOperations.ts
libera13: issue for scheduling https://github.com/naira-project/naira/issues/157

ui-poc/src/App.tsx
kocakkaan: Yes, I agree. For the initial implementation however, I think being generic would be better. After this PR, we can think about which parts to make as a component (such as a tabular view which we see in every viewpoint). I would also include this thought into ticket #103 .

rootlsz (approved):
i;m not a react expert but I have a feeling it's nice :D  + I learned something new about taskfile :D (the underscore notation) 👏

ui-poc/src/hooks/useKinds.ts
```diff
@@ -14,45 +16,41 @@ interface UseKindsResult {
...
-      })
-      .catch(() => {
-        setKindsError('Failed to discover kinds');
-        setKindsLoading(false);
-      });
-  }, [allowedKinds?.join(','), token]);
+  const kinds = useMemo(
+    () => (allowedKinds ? allowedKinds.filter((k) => discoveredKinds.includes(k)) : discoveredKinds),
+    [allowedKinds?.join(','), discoveredKinds],
+  );
```
Daviidg, line 37: Not 100% sure where you get the allowedKinds, but shouldnt the backend do this filtering? 

If the user isnt allowed to see certain kinds, its usually better to handle it on the backend and not allow the request to reach FE

kocakkaan: I realized this is a bad naming, here it is just hardcoded kinds and plugins per viewpoint. It is not related to any authorization mechanism. You can see these hardcoded points in `src/config/viewpoints.ts` in this PR. I am changing the naming now. Thanks 👍

ui-poc/src/components/PluginsManagerDialog.tsx
kocakkaan: Created a new Plugins Page for this purpose. And now the plugin names are shown in the empty state.

PR #154, opened 2026-08-18
feat: add e2e test infra with smoke test

### LINKED ISSUE #181: Implement E2E Test infra and run it on pull request
Implement the E2E test infrastructure and integrate E2E test execution into the pull request (PR) workflow.

The decision to implement this approach was documented and agreed upon in the following RFC:

- https://github.com/naira-project/architecture/pull/19

This issue covers the implementation of the E2E testing infrastructure and the required CI integration to automatically execute the tests for pull requests.

Scope

- Implement the E2E test infrastructure as defined in the RFC.
- Integrate E2E test execution into the PR CI workflow.
- Ensure the test environment is provisioned and configured automatically.
- Ensure test results are reported as part of the PR workflow.
- Clean up test resources after execution where applicable.

Acceptance Criteria

- [ ] E2E test infrastructure is implemented according to the RFC.
- [ ] E2E tests can be executed automatically for pull requests.
- [ ] Required test dependencies and configuration are provisioned automatically.
- [ ] E2E test results are clearly visible in the PR workflow.
- [ ] Test resources are cleaned up after execution.
- [ ] Documentation is updated with instructions for running and troubleshooting E2E tests locally and in CI.

jagdeepm:
## Description

Adds E2E test infrastructure for Naira: a shared component pool under e2e/components/ (catalog + plugin sidecars, keycloak, litellm, mlflow, llamacpp, openmetadata, postgres, ui, portal, chatbot1) that scenarios opt into piece-by-piece via a components.env file, plus one scenario — litellm_chatbot_to_catalog_api — that proves the litellm and depl_uses_litellm catalog plugins merge into one subgraph, on a throwaway kind cluster.

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

Successful CI test run on workflow-testing repo 
https://github.com/naira-project/workflow-testing/actions/runs/33858285572/job/100976552827

- [ ] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

Used Claude code to generate code and review


rootlsz (commented):
I'm thinking about a correct approach for reviewing this pr, maybe you can help me with that.

There are lots of yaml and configs, and only a single `smoke_test.go` file. 

I can't find #154 under github Actions, don't know if I can do anything about it (how do I review github configs ⁠:D)

Test are visible in go file 👍 

I assume it is a solid core that we can extend in the future? `Naira Core test stratergy` does not specify any details about smoke tests 

There are API tests listed in `Naira Core API test cases` document, but only for UI. 

If it's only infra and we don't need to write them down in some document, it's fine.

deploy/e2e/README.md
```diff
@@ -0,0 +1,98 @@
+# E2E Test Environments
+
+Spins up the entire Naira stack — catalog, portal, UI, MLflow, LiteLLM, llama.cpp,
```
akavel-reply: As explained in https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation , for e2e test described there, I don't want all plugins, just 2 of them - so, I don't want "entire" Naira stack (which would require all plugins).

Also, MLflow, LiteLLM, llama.cpp are _not_ part of Naira stack, so you must not describe them as such.

deploy/e2e/README.md
```diff
@@ -0,0 +1,98 @@
...
+Keycloak, OpenMetadata — into one fresh, isolated Kubernetes namespace, seeds it
+with known data, and runs smoke tests against it. Used per-PR in CI; also
```
akavel-reply: "tests" - misleading and incorrect: plural noun but there's only one test as of now

deploy/e2e/README.md
```diff
@@ -0,0 +1,98 @@
...
+runnable locally before pushing.
+
+
+## Prerequisites
+
+`kind`, `kubectl`, `helm`, `docker`, `envsubst`, `go` — pinned versions in the
+repo's `mise.toml`. Use `mise` to pick these up automatically; a `kind` older
```
akavel-reply: fix or explain: above looks untrue, I don't see those tools in mise.toml

deploy/e2e/README.md
```diff
@@ -0,0 +1,98 @@
+# E2E Test Environments
+
+Spins up the entire Naira stack — catalog, portal, UI, MLflow, LiteLLM, llama.cpp,
+Keycloak, OpenMetadata — into one fresh, isolated Kubernetes namespace, seeds it
```
akavel-reply: As explained in https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation , for e2e test described there, I don't want all plugins, just 2 of them - so, I don't want "entire" Naira stack (which would require all plugins).

Also, as described there, I don't want OpenMetadata nor llama.cpp to be running in this test scenario.

Also, MLflow, LiteLLM, llama.cpp, OpenMetadata are _not_ part of Naira stack, so you must not describe them as such.

deploy/e2e/README.md
```diff
@@ -0,0 +1,98 @@
...
+./deploy/e2e/scripts/destroy-environment.sh --env-id "$NS" --cluster-name naira-idp-e2e
+kind delete cluster --name naira-idp-e2e
+```
+
+## What's in the namespace
+
+| Component | Notes |
+|---|---|
+| `catalog` + 6 plugin sidecars | Reads MLflow/OpenMetadata/LiteLLM/K8s into the graph |
+| `portal`, `ui` | Frontend |
```
akavel-reply: Per our earlier discussion, we agreed to not test UI in this scenario, so you must not start UI.

deploy/e2e/README.md
```diff
@@ -0,0 +1,98 @@
...
+| `keycloak` | Auth — seeded with a `naira-portal` client and a `testuser`/`testpass` user |
+| `mlflow`, `postgres`, `litellm`, `llama-dummy-model` | AI infra, one demo model registered |
+| OpenMetadata (Helm: server + MySQL + OpenSearch) | 5 demo tables + lineage seeded |
```
akavel-reply: You must remove this, it is not used for test described in https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation Therefore, starting this is misleading for the test, wastes resources, increases complexity unnecessarily, increases chances of accidental interactions, and wastes cognitive load of engineers maintaining and managing this test in the future. We want the simplest environment possible for given test, so when test verifies only 2 plugins, we must use only those components that are necessary for those 2 plugins.

Note the following text in above linked page:

> - no extra stuff is running within the (cluster/)namespace

deploy/e2e/README.md
```diff
@@ -0,0 +1,98 @@
...
+## Infra capacity needed
+
+Per PR/run, enforced by the namespace's `ResourceQuota`
+(`deploy/e2e/quota.yaml`): **~3.5 CPU / ~8.5Gi memory requested**, up to
+**11 CPU / 15Gi memory** at burst (limits). OpenMetadata (server + MySQL +
+OpenSearch) is over half of that on its own. PVCs declare ~36Gi, but on
+`kind` (hostPath-backed `local-path-provisioner`) actual disk used is only
+~1Gi — a cloud cluster with real dynamic provisioning would actually
+allocate the full 36Gi. See the SPEC's "Resource Cost" section for the
+per-component breakdown.
```
akavel-reply: What is "SPEC"? where do I find it? You must clarify to make it easier to understand and find for person reading this.

tests/e2e/smoke_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+// Package e2e holds smoke tests that run against a live, seeded environment
+// created by deploy/e2e/scripts/create-environment.sh. Excluded from the
+// default `go test ./...` build via the e2e build tag — run explicitly with
+// `go test -tags e2e ./tests/e2e/...`.
+//
+// The `full` deployment profile means every dependency is empty except for
+// what the seed Job put there, so these assertions can be exact (a specific
+// count, a specific node) rather than mere existence checks. See the RFC's
+// "Isolation Scope" for why that distinction matters.
```
akavel-reply: I don't know what RFC - you must link specific RFC or at least give its number

tests/e2e/smoke_test.go
```diff
@@ -0,0 +1,160 @@
...
+func authToken(t *testing.T) string {
+	t.Helper()
+	authTokenOnce.Do(func() {
+		keycloakURL := envOr("KEYCLOAK_URL", "http://localhost:8080")
+		form := url.Values{
+			"grant_type":    {"password"},
+			"client_id":     {envOr("KEYCLOAK_CLIENT_ID", "naira-portal")},
+			"client_secret": {envOr("KEYCLOAK_CLIENT_SECRET", "naira-e2e-test-secret")},
+			"username":      {envOr("KEYCLOAK_USERNAME", "testuser")},
+			"password":      {envOr("KEYCLOAK_PASSWORD", "testpass")},
```
akavel-reply: why `envOr` here? do we use it anywhere? if no, please remove, we must not create dead code

tests/e2e/smoke_test.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		tokenURL := fmt.Sprintf("%s/realms/%s/protocol/openid-connect/token", keycloakURL, envOr("KEYCLOAK_REALM", "naira"))
```
akavel-reply: again, remove `envOr` if not used

tests/e2e/smoke_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("GET /healthz: status %d, want 200", resp.StatusCode)
+	}
+}
+
+// The MLflow seed script (register_dummy_model.py) registers exactly one
+// model named "demo-model" under the "mlflow" plugin path prefix, into an
+// MLflow instance that — under the `full` profile — belongs to this PR alone.
+func TestMLflowModelSeeded(t *testing.T) {
```
akavel-reply: I did not ask for MLFlow test in https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation , please remove this test

tests/e2e/smoke_test.go
```diff
@@ -0,0 +1,160 @@
...
+		t.Fatalf("nodes matching mlflow/demo-model = %d, want exactly 1 (got %+v)", len(got.Nodes), got.Nodes)
+	}
+	if got.Nodes[0].Kind != "model" {
+		t.Errorf("node kind = %q, want %q", got.Nodes[0].Kind, "model")
+	}
+}
+
+// seed_sample_tables.py creates exactly 5 tables. Under `full` this OpenMetadata
+// instance is exclusively this PR's, so the count is exact — not "at least 5".
+func TestOpenMetadataTablesSeeded(t *testing.T) {
```
akavel-reply: I did not ask for OpenMetadata test in https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation , please remove this test

tests/e2e/smoke_test.go
akavel-reply, line 1: MAJOR: From perspective of what I described in https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation , this file is totally not what I want. I described "ASSERT" section in the linked document - I don't see it here:

- that section describes what I want test verification/assertion to do - but I don't see its implementation here; is it somewhere else? **correct implementation of that document's "ASSERT" section is a blocking requirement for me to accept a PR as implementation of that test** - it's part of "acceptance criteria" for me
- this file implements some "tests" that I am completely uninterested in from perspective of https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation , I don't want them - please remove them.

deploy/e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env bash
+# Creates a full per-PR E2E environment on an existing kind cluster: builds and
+# loads every image, deploys the whole platform into a fresh namespace, waits
+# for it to be genuinely ready, and seeds a starting dataset.
+#
+# Only --profile full is implemented. See SPEC-Kubernetes-Test-Environments-Implementation.md
```
akavel-reply: where is this SPEC file? I don't see it

deploy/e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,170 @@
...
+#
+# Usage:
+#   create-environment.sh --env-id <ID> --tag <TAG> [--cluster-name <NAME>] [--profile full]
+#   create-environment.sh --pr <N> --tag <TAG> [--run-id <ID>] [--cluster-name <NAME>] [--profile full]
+#
+# --env-id is preferred: pass an ID the caller already computed (e.g. the
+# workflow, before this script runs) so a teardown step can find the right
+# environment even if this script fails before it would otherwise have
+# derived one. --pr/--run-id remain for local/manual convenience.
+#
```
akavel-reply: explain clearly all the flags in a standard way like in `--help` - they are not explained now, I don't understand what they do:

```
--env-id ID - <explanation...>
--tag TAG - <explanation...>
<etc...>
```

akavel-reply (changes requested):
**I am working under assumption that this is intended to implement what we discussed and agreed on in: https://liquid-naira.atlassian.net/wiki/spaces/Naira/pages/123633665/E2E+integration+test+scenario+s+ideation If this assumption is wrong, help me understand why.**

Under above assumption, the PR currently **does not implement** what I described in above link. Major discrepancies:

1. The PR starts **unnecessary services and components**:
    - too many plugins - above document requires exactly 2 plugins (named in the document), this PR runs different plugins
    - unnecessary components are started: MLFlow, OpenMetadata, llama.cpp - none of them are described in document linked above
    - in `tests/e2e/smoke_test.go`, there are unnecessary tests not requested by document linked above
2. The PR appears to be **missing crucial components** described in document linked above:
    - "chatbot1" - see description in document linked above
    - "ASSERT" - the document describes very concrete script and assertion required for tests to pass successfully - I don't see clearly where it is implemented, seems to be missing

**In addition**, I request change of structure of directories from this PR for a test as described in above document:

1. Move all contents into directory `TEST_ROOT`, defined as: **`REPO/e2e/litellm_chatbot_to_catalog_api/`
2. Within above `TEST_ROOT`, suggested files and subdirectories:
```
README.md - must be in base TEST_ROOT dir
scripts/ - contents can stay as they are now
base/ (or platform/) - base platform infrastructure before starting components of test - kind-config.yaml, quota.yaml
components/ - core
seed/ - contents can stay as they are now, PLUS: seed-job.yaml
assert/ - any verification code deciding on PASS or FAIL result of test - can be in Go, Python, bash, or whatever else fits best (see "ASSERT" section of document linked above)
```

I **also** added some other comments in the PR related to other aspects of quality that need to also be addressed.

**MAKE SURE TO REPLY TO EVERY REVIEW COMMENT RAISED IN THIS PR.** Either reply "implemented", or explain why you think it is better to make a different decision.

**PING ME PERSONALLY ON Teams WHEN PR IS READY FOR NEXT ROUND OF REVIEW FROM ME.** I encourage you to also call me earlier with any questions and/or unclarities you have related to this topic.

deploy/e2e/README.md
jagdeepm: I have refactored the code now to satisfy the test scenario you have described it does not install all the components now

deploy/e2e/README.md
jagdeepm: Refactored the docs

deploy/e2e/README.md
jagdeepm: updated the docs

deploy/e2e/README.md
jagdeepm: Updated test scenarios and docs

deploy/e2e/README.md
jagdeepm: There are no UI tests as of now

deploy/e2e/README.md
jagdeepm: Removed

deploy/e2e/README.md
jagdeepm: Removed

tests/e2e/smoke_test.go
jagdeepm: Removed

tests/e2e/smoke_test.go
jagdeepm: code removed

tests/e2e/smoke_test.go
jagdeepm: code removed

tests/e2e/smoke_test.go
jagdeepm: Removed test code

tests/e2e/smoke_test.go
jagdeepm: Removed test code

tests/e2e/smoke_test.go
jagdeepm: Test structure address this

deploy/e2e/scripts/create-environment.sh
jagdeepm: Removed

deploy/e2e/scripts/create-environment.sh
jagdeepm: Added usage instructution to script

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+// authenticates against litellm1 — see components/chatbot1.yaml) and emits
+// exactly one uses_model edge to the "openai" model it's configured to use.
+func TestChatbotUsesLitellmModel(t *testing.T) {
+	filter := url.QueryEscape(`kind="uses_model"`)
+	var got listRelationsResponse
+	getJSON(t, fmt.Sprintf("/v1/relations?filter=%s", filter), &got)
+
+	if len(got.Relations) != 1 {
+		t.Fatalf("uses_model relations = %d, want exactly 1 (got %+v)", len(got.Relations), got.Relations)
+	}
```
akavel-reply: In the tests in Naira, we use the `assert` and `require` subpackages of the open-source `github.com/stretchr/testify` library. Please use `require.Len()` here:


```suggestion
	require.Len(t, got.Relations, 1)
```

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+	rel := got.Relations[0]
+	if !strings.HasSuffix(rel.FromNode, "/chatbot1") {
+		t.Errorf("fromNode = %q, want it to end in %q", rel.FromNode, "/chatbot1")
+	}
+	if rel.ToNode != "nodes/model/litellm1/openai" {
+		t.Errorf("toNode = %q, want %q", rel.ToNode, "nodes/model/litellm1/openai")
+	}
```
akavel-reply: Similar as in comment above, please use `assert.Equal()` here:


```suggestion
	assert.Equal(t, "nodes/model/litellm1/openai", rel.ToNode)
```

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+	var got listRelationsResponse
+	getJSON(t, fmt.Sprintf("/v1/relations?filter=%s", filter), &got)
+
+	if len(got.Relations) != 1 {
+		t.Fatalf("uses_model relations = %d, want exactly 1 (got %+v)", len(got.Relations), got.Relations)
+	}
+	rel := got.Relations[0]
+	if !strings.HasSuffix(rel.FromNode, "/chatbot1") {
+		t.Errorf("fromNode = %q, want it to end in %q", rel.FromNode, "/chatbot1")
+	}
```
akavel-reply: Similar as in comment above and below, for consistency please use `assert.Truef()` here:


```suggestion
	assert.Truef(t, strings.HasSuffix(rel.FromNode, "/chatbot1"),
		"fromNode = %q, want it to end in %q", rel.FromNode, "/chatbot1")
```

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("GET /healthz: status %d, want 200", resp.StatusCode)
+	}
+}
+
+// The litellm plugin discovers every model litellm.yaml's static config
+// lists (see components/litellm.yaml) — exactly "openai" and "mistral",
+// under the litellm1 PATH_PREFIX both plugins in this scenario share.
+func TestLitellmModelsSeeded(t *testing.T) {
```
akavel-reply: This test doesn't just check if the models are seeded (that would only check in LiteLLM's API), but rather properly checks in the catalog API. The name of the test needs to be adjusted, for example:


```suggestion
func TestLitellmModelsNodes(t *testing.T) {
```

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+	filter := url.QueryEscape(`kind="model"`)
+	var got listNodesResponse
+	getJSON(t, fmt.Sprintf("/v1/nodes?filter=%s", filter), &got)
+
+	want := map[string]bool{"litellm1/openai": true, "litellm1/mistral": true}
+	if len(got.Nodes) != len(want) {
+		t.Fatalf("model nodes = %d, want exactly %d (got %+v)", len(got.Nodes), len(want), got.Nodes)
+	}
```
akavel-reply: Similar as suggested in comments below, please use `assert.Len()` here:


```suggestion
	assert.Len(t, got.Nodes, len(want))
```

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+	for _, n := range got.Nodes {
+		if !want[n.Path] {
+			t.Errorf("unexpected model node path %q", n.Path)
+			continue
+		}
+		delete(want, n.Path)
+	}
+	for path := range want {
+		t.Errorf("missing model node path %q", path)
+	}
```
akavel-reply: This code seems correct, but to make it a bit simpler and easier to understand, please rewrite as below:

```
var response listNodesResponse
getJSON(t, fmt.Sprintf("/v1/nodes?filter=%s", filter), &response)

wantPaths := []string{"litellm1/openai", "litellm1/mistral"}
var paths []string
for _, n := range response.Nodes {
	paths = append(paths, n.Path)
}
assert.ElementsMatchf(t, wantPaths, paths, "full response: %#v", response)
```

e2e/litellm_chatbot_to_catalog_api/README.md
```diff
@@ -0,0 +1,118 @@
...
+then asserts the resulting graph via the catalog API. Used per-PR in CI;
+also runnable locally before pushing. See [`e2e/README.md`](../README.md)
+for how this scenario's component selection works and how to add another
+scenario.
+
+## Prerequisites
+
+`kind`, `kubectl`, `helm`, `go`, `python` — pinned versions in the repo's
+`mise.toml`; use `mise` to pick these up automatically. `docker` and
+`envsubst` are not mise-managed — install them separately (Docker
```
akavel-reply: Why `envsubst` is not mise-managed? We use mise specifically to have deterministic environment, so I'd recommend actually adding it into mise.toml, unless there are important reasons against.

As to docker, I'd however still keep it outside as is, as its licensing situation is complicated.

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+RUN_ID="$(date +%s)"
+PR=""
+TAG=""
+ENV_ID=""
+SCENARIO=""
+
+usage() {
+  cat >&2 <<'EOF'
+Usage: create-environment.sh --scenario <NAME> --env-id <ID> --tag <TAG> [--cluster-name <NAME>]
+       create-environment.sh --scenario <NAME> --pr <N> --tag <TAG> [--run-id <ID>] [--cluster-name <NAME>]
```
akavel-reply: This variant seems unused (from what I see, PR workflow seems to just create its ENV_ID manually). If so, let's delete it and its flags to avoid dead code.

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  docker build -q -t "${image}:${TAG}" -f "${dockerfile_dir}/Dockerfile" "${REPO_ROOT}" >/dev/null
+  kind load docker-image "${image}:${TAG}" --name "${CLUSTER_NAME}"
+}
+
+build_and_load_self_context() {
+  local image="$1" dir="$2"
+  echo "==> Building ${image}:${TAG}"
+  docker build -q -t "${image}:${TAG}" "${dir}" >/dev/null
+  kind load docker-image "${image}:${TAG}" --name "${CLUSTER_NAME}"
+}
```
akavel-reply, line 131: those two funcs look like they could be simplified to just one, taking 2 mandatory args plus 1 optional arg:
```
local image="$1"
local dockerfile_dir="$2"
local build_root="${3:-$2}"
```
Then, for the "self-context" calls, pass $REPO_ROOT as the third arg.

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  local image="$1" dockerfile_dir="$2"
+  echo "==> Building ${image}:${TAG}"
+  docker build -q -t "${image}:${TAG}" -f "${dockerfile_dir}/Dockerfile" "${REPO_ROOT}" >/dev/null
+  kind load docker-image "${image}:${TAG}" --name "${CLUSTER_NAME}"
+}
+
+build_and_load_self_context() {
+  local image="$1" dir="$2"
+  echo "==> Building ${image}:${TAG}"
+  docker build -q -t "${image}:${TAG}" "${dir}" >/dev/null
```
akavel-reply: why redirect to null? what if a docker build fails?

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  kind load docker-image "${image}:${TAG}" --name "${CLUSTER_NAME}"
+}
+
+echo "==> Building and loading images"
+build_and_load catalog "${REPO_ROOT}/catalog"
+build_and_load_self_context seed "${SCENARIO_DIR}/seed"
+for plugin in ${PLUGINS}; do
+  build_and_load "${plugin}" "${REPO_ROOT}/$(plugin_image_dir "${plugin}")"
```
akavel-reply: It should be possible to delete `plugin_image_dir` func, and simplify this to:


```suggestion
  build_and_load "${plugin}" "${REPO_ROOT}/plugins/cmd/${plugin//-/_}"
```

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  for plugin in ${PLUGINS}; do
+    port="$(plugin_port "${plugin}")"
+    addresses="${addresses:+${addresses},}${plugin}=localhost:${port}"
+  done
+  export PLUGIN_ADDRESSES="${addresses}"
+
+  {
+    cat "${E2E_DIR}/components/catalog/header.yaml"
+    for plugin in ${PLUGINS}; do
+      cat "${E2E_DIR}/components/catalog/plugin-${plugin}.yaml"
```
akavel-reply: You could generate incrementing port number in the loop, and envsubst it here. Then, you could remove the whole `plugin_port` function and we wouldn't have to remember to maintain it manually.

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  build_dir="$(component_build_dir "${component}")"
+  if [ -n "${build_dir}" ]; then
+    build_and_load_self_context "${component}" "${build_dir}"
+  fi
+done
+
+# ---------------------------------------------------------------------------
+# Deploy. catalog first — it creates the Namespace every other manifest
+# depends on. quota.yaml right after, before anything else lands, so nothing
+# is ever deployed unmetered even transiently.
```
akavel-reply: Why not create quota and keycloak first? who said Namespace must be created within `catalog.yaml` and cannot be moved elsewhere? Current order is confusing, I would expect we want catalog & keycloak also created under quotas already.

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  if [ "${component}" = "openmetadata" ]; then
+    echo "==> Deploying OpenMetadata (Helm) — this is the long pole, 5-9 min"
+    helm repo add open-metadata https://helm.open-metadata.org/ --force-update
+    helm repo update open-metadata
+    apply "${E2E_DIR}/components/openmetadata-secrets.yaml"
+    helm upgrade --install openmetadata-dependencies open-metadata/openmetadata-dependencies \
+      --namespace "${NAMESPACE}" --values "${E2E_DIR}/components/openmetadata-deps-values.yaml" --wait --timeout 15m
+    helm upgrade --install openmetadata open-metadata/openmetadata \
+      --namespace "${NAMESPACE}" --values "${E2E_DIR}/components/openmetadata-values.yaml" --wait --timeout 15m
+    continue
```
akavel-reply, line 194: Do we specify concrete, pinned versions of openmetadata helm charts? if yes, where? if not, why?

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+    *) echo "$1" ;;
+  esac
+}
+
+component_wait_timeout() {
+  case "$1" in
+    litellm) echo "900s" ;;
+    llamacpp) echo "600s" ;;
+    openmetadata) echo "600s" ;;
+    *) echo "180s" ;;
```
akavel-reply: those all seem to be full minutes intervals - can we specify them in minutes for easier reading and understanding?

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+    portal) echo "${REPO_ROOT}/naira-openmfp-portal" ;;
+    chatbot1) echo "${E2E_DIR}/components/chatbot1" ;;
+    litellm|mlflow|llamacpp|openmetadata|postgres) echo "" ;;
+    *) echo "error: unknown component '$1'" >&2; exit 1 ;;
+  esac
+}
+
+component_deployment() {
+  case "$1" in
+    llamacpp) echo "llama-dummy-model" ;;
```
akavel-reply: Why this single special-case? can't we just remove this func, and make sure we use a single, consistent naming for everything _including_ the llamacpp deployment with dummy model?

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  apply "${E2E_DIR}/components/${component}.yaml"
+done
+
+# ---------------------------------------------------------------------------
+# Readiness. Each rollout status wait only succeeds once the Deployment's own
+# readinessProbe passes — no fixed sleeps anywhere in this script.
+# ---------------------------------------------------------------------------
+echo "==> Waiting for dependency rollouts"
+kubectl -n "${NAMESPACE}" rollout status deploy/keycloak --timeout=300s
+for component in ${COMPONENTS}; do
```
akavel-reply: looks like we could do:

```suggestion
for component in keycloak ${COMPONENTS}; do
```

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+  kubectl -n "${NAMESPACE}" rollout status "deploy/$(component_deployment "${component}")" \
+    --timeout="$(component_wait_timeout "${component}")"
```
akavel-reply, line 206: (optional) instead of defining `component_wait_timeout` function, we could specify timeouts in comments in the YAML files, and use grep or sed to extract them here, defaulting to 180s.

For example, in `litellm.yaml`, we could put a comment on top of the file:
```
# ROLLOUT_TIMEOUT=15m
```

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,276 @@
...
+done
+kubectl -n "${NAMESPACE}" rollout status deploy/catalog --timeout=180s
```
akavel-reply: This could probably also be merged into the `for` loop:

```bash
for component in keycloak ${COMPONENTS} catalog; do
```

e2e/litellm_chatbot_to_catalog_api/seed/provision_chatbot1_key.py
```diff
@@ -0,0 +1,71 @@
...
+Secrets in this namespace, see seed-job.yaml) to patch the Secret directly —
+stdlib-only, no Kubernetes client library needed for one PATCH call.
+"""
+
+import json
+import os
+import ssl
+import urllib.request
+
+LITELLM_URL = os.environ.get("LITELLM_URL", "http://localhost:4000").rstrip("/")
```
akavel-reply: Do we use this `localhost` for anything? can we delete it?

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+			"client_secret": {keycloakClientSecret},
+			"username":      {keycloakUsername},
+			"password":      {keycloakPassword},
+		}
+		tokenURL := fmt.Sprintf("%s/realms/%s/protocol/openid-connect/token", keycloakURL, keycloakRealm)
+
+		resp, err := httpClient().Post(tokenURL, "application/x-www-form-urlencoded", strings.NewReader(form.Encode()))
+		if err != nil {
+			t.Fatalf("POST %s: %v", tokenURL, err)
+		}
```
akavel-reply: please shorten to `require.NoError(t, err)` (as is standard in our tests), and other helper funcs from `testify/assert` and `testify/require` libraries - here and in other similar places in this file

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+	Relations []relation `json:"relations"`
+	TotalSize int32      `json:"totalSize"`
+}
+
+func catalogURL(t *testing.T) string {
+	t.Helper()
+	if u := os.Getenv("CATALOG_URL"); u != "" {
+		return u
+	}
+	return "http://localhost:8090"
```
akavel-reply: I don't see a reason to need this - let's remove the func and just use os.Getenv

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func httpClient() *http.Client {
+	return &http.Client{Timeout: 10 * time.Second}
+}
+
+func envOr(key, fallback string) string {
```
akavel-reply: don't see a reason to need this - let's remove the func and just use os.Getenv

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("GET %s: status %d", fullURL, resp.StatusCode)
+	}
+	if err := json.NewDecoder(resp.Body).Decode(out); err != nil {
+		t.Fatalf("GET %s: decoding response: %v", fullURL, err)
+	}
+}
+
+func TestHealthz(t *testing.T) {
```
akavel-reply: Can we remove this test?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,155 @@
...
+            sleep 1
+          done
+          echo "::error::keycloak not reachable via port-forward after 30s" >&2
+          exit 1
+
+      - name: Run E2E tests
+        env:
+          CATALOG_URL: http://localhost:8090
+          KEYCLOAK_URL: http://localhost:8080
+        run: go test -tags e2e -v ./e2e/litellm_chatbot_to_catalog_api/assert/...
```
akavel-reply: Is this retrying with some timeout? If not, why? if yes, where's the timeout defined?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,155 @@
...
+
+      - name: Stop port-forwards
+        if: always()
+        run: |
+          kill "$(cat /tmp/catalog-port-forward.pid)" 2>/dev/null || true
+          kill "$(cat /tmp/keycloak-port-forward.pid)" 2>/dev/null || true
+
+      - name: Collect diagnostics
+        if: failure()
```
akavel-reply: Can we show diagnostics also in case of success, to allow manual sanity-check review?

e2e/litellm_chatbot_to_catalog_api/seed/trigger_catalog_sync.py
```diff
@@ -0,0 +1,85 @@
...
+import urllib.parse
+import urllib.request
+
+CATALOG_URL = os.environ.get("CATALOG_URL", "http://localhost:8090").rstrip("/")
+KEYCLOAK_URL = os.environ.get("KEYCLOAK_URL", "http://localhost:8080").rstrip("/")
+KEYCLOAK_REALM = os.environ.get("KEYCLOAK_REALM", "naira")
+KEYCLOAK_CLIENT_ID = os.environ.get("KEYCLOAK_CLIENT_ID", "naira-portal")
+KEYCLOAK_CLIENT_SECRET = os.environ.get("KEYCLOAK_CLIENT_SECRET", "naira-e2e-test-secret")
+KEYCLOAK_USERNAME = os.environ.get("KEYCLOAK_USERNAME", "testuser")
+KEYCLOAK_PASSWORD = os.environ.get("KEYCLOAK_PASSWORD", "testpass")
```
akavel-reply: Do we need all those alternative values? It looks like we define some of them in the job, so what's the point to use default values? On the other hands, some we don't seem to define in the job - so why read them from environ at all?

e2e/components/postgres.yaml
```diff
@@ -0,0 +1,99 @@
...
+    matchLabels:
+      app: postgres
+  template:
+    metadata:
+      labels:
+        app: postgres
+    spec:
+      containers:
+        - name: postgres
+          image: postgres:16-alpine
```
akavel-reply: It would be better to use precise version, for better determinism.

akavel-reply (changes requested):
General shape of the thing looks much more fitting for me now, thanks for the update!

I added a bunch of smaller comments.

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Done — updated to require.Len(t, got.Relations, 1).

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Done — updated to assert.Equal()

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Done — updated to assert.Truef().

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Renamed to TestLitellmModelsNodes.

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Rewritten as suggested.

e2e/scripts/create-environment.sh
jagdeepm: Merged into one build_and_load(image, dockerfile_dir, build_root=dockerfile_dir) as suggested.

e2e/scripts/create-environment.sh
jagdeepm: Removed — only documents the real --scenario/--env-id/--tag/--cluster-name flags now.

e2e/scripts/create-environment.sh
jagdeepm: Removed the redirect — build output isn't suppressed, and set -euo pipefail already fails the script on a build error.

e2e/scripts/create-environment.sh
jagdeepm: Removed plugin_image_dir — build_and_load is called directly with the plugin's cmd dir now.

e2e/scripts/create-environment.sh
jagdeepm: Removed plugin_port — ports are generated inline in the loop now.

e2e/scripts/create-environment.sh
jagdeepm: Reordered — namespace, then quota.yaml, then keycloak.yaml, then catalog last.

e2e/scripts/create-environment.sh
jagdeepm: Pinned to --version 1.13.5 for both OpenMetadata charts.

e2e/scripts/create-environment.sh
jagdeepm: Already in minutes (15m/10m/3m) after the component_wait_timeout rewrite.

e2e/scripts/create-environment.sh
jagdeepm: Removed component_deployment — renamed llamacpp's Deployment/Service from llama-dummy-model to llamacpp so no name mapping is needed.

e2e/scripts/create-environment.sh
jagdeepm: Folded into the single readiness loop below.

e2e/scripts/create-environment.sh
jagdeepm: Done — for component in keycloak ${COMPONENTS} catalog; do, one loop.

e2e/litellm_chatbot_to_catalog_api/seed/provision_chatbot1_key.py
jagdeepm: Removed — LITELLM_URL is a required env var now, no localhost fallback.

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Removed — getJSON reads os.Getenv("CATALOG_URL") directly now.

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Removed — authToken reads os.Getenv("KEYCLOAK_URL") directly now.

.github/workflows/e2e.yml
jagdeepm: Added -timeout=5m to the go test invocation, plus a comment explaining why.

.github/workflows/e2e.yml
jagdeepm: Changed Collect diagnostics to if: always().

e2e/litellm_chatbot_to_catalog_api/seed/trigger_catalog_sync.py
jagdeepm: Removed the optional/default-valued env vars — realm/client/user are hardcoded constants now; CATALOG_URL/KEYCLOAK_URL are required (os.environ[...], no fallback).

e2e/components/postgres.yaml
jagdeepm: Pinned to postgres:16.4-alpine.

e2e/litellm_chatbot_to_catalog_api/assert/assert_test.go
jagdeepm: Removed

e2e/scripts/create-environment.sh
jagdeepm: skipped for now , can be considered later

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,165 @@
...
+          chmod +x ./kubectl
+          sudo mv ./kubectl /usr/local/bin/kubectl
+
+      - name: Install helm
+        run: |
+          curl -Lo ./helm.tar.gz "https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz"
+          curl -Lo ./helm.tar.gz.sha256 "https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz.sha256sum"
+          echo "$(cut -d' ' -f1 ./helm.tar.gz.sha256)  ./helm.tar.gz" | sha256sum -c -
+          tar -xzf ./helm.tar.gz
+          sudo mv ./linux-amd64/helm /usr/local/bin/helm
```
akavel-reply, line 81: Is there a reason the PR doesn't use `mise` to fetch those tools? we already have it setup in the project, introducing a separate mechanism for fetching increases complexity and a risk of those versions desynchronizing.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,165 @@
...
+        env:
+          CATALOG_URL: http://localhost:8090
+          KEYCLOAK_URL: http://localhost:8080
+        run: go test -tags e2e -v -timeout=5m ./e2e/litellm_chatbot_to_catalog_api/assert/...
+
+      - name: Stop port-forwards
+        if: always()
+        run: |
+          kill "$(cat /tmp/catalog-port-forward.pid)" 2>/dev/null || true
+          kill "$(cat /tmp/keycloak-port-forward.pid)" 2>/dev/null || true
```
akavel-reply, line 148: why the `2>/dev/null` ?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,165 @@
...
+
+      - name: Collect diagnostics
+        if: always()
+        run: |
+          kubectl get pods,services -n "${ENV_ID}" -o wide || true
+          kubectl logs -n "${ENV_ID}" -l app=catalog --all-containers --tail=-1 || true
```
akavel-reply, line 154: please add `--prefix=true`

.github/workflows/README.md
```diff
@@ -24,4 +24,37 @@ Ensures code quality and buildability. Gatekeeper: `All Checks Passed`.
...
+for this scenario's exact run steps:
+
+```bash
+# 1. Create the cluster
+kind create cluster --config e2e/base/kind-config.yaml
+
+# 2. Build, load, deploy, and seed this scenario into a fresh namespace
+./e2e/scripts/create-environment.sh --scenario litellm_chatbot_to_catalog_api \
+  --env-id local --tag localtest --cluster-name naira-idp-e2e
+# → prints ENV_ID=<namespace> on success, note it for step 3
```
akavel-reply, line 45: From what I see, the script just prints the same value as was passed in the `--env-id local` flag, no? so, it will always print `ENV_ID=local` here, no? From this it seems to me we don't need to "note it for step 3", we can just specify `NS=local` there, no?

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,221 @@
...
+  echo "==> Rendering catalog (plugins: ${PLUGINS:-<none>})"
+  local addresses="" plugin port=50051
+  for plugin in ${PLUGINS}; do
+    addresses="${addresses:+${addresses},}${plugin}=localhost:${port}"
+    port=$((port + 1))
+  done
+  export PLUGIN_ADDRESSES="${addresses}"
+
+  {
+    envsubst '${NAMESPACE} ${ENV_ID} ${TAG} ${PLUGIN_ADDRESSES}' < "${E2E_DIR}/components/catalog/header.yaml"
```
akavel-reply, line 168: From earlier code, it seems NAMESPACE and ENV_ID are always equal. Is there any reason why we need to pass ENV_ID into the `envsubst` calls? I would expect it enough to just pass NAMESPACE (or anyway just one of them, regardless which).

e2e/components/chatbot1/app.py
```diff
@@ -0,0 +1,52 @@
...
+depl_uses_litellm has a real Deployment to discover: it reads LITELLM_API_KEY
+from a Secret, matching that plugin's API-key-in-a-Secret detection.
+"""
+
+import os
+
+import requests
+import streamlit as st
+
+LITELLM_API_URL = os.environ.get("LITELLM_API_URL", "http://localhost:4000")
```
akavel-reply, line 15: Why do we need the `localhost:4000` value here?

e2e/components/chatbot1/app.py
```diff
@@ -0,0 +1,52 @@
...
+LITELLM_API_KEY = os.environ.get("LITELLM_API_KEY", "")
+MODEL = os.environ.get("LITELLM_MODEL", "openai")
```
akavel-reply, line 17: why do we need the two alternative values (env or default) here? I would expect only one of them

e2e/components/chatbot1.yaml
```diff
@@ -0,0 +1,74 @@
...
+                  name: chatbot1-secrets
+                  key: LITELLM_API_KEY
+          ports:
+            - containerPort: 8501
+              name: http
+          readinessProbe:
+            httpGet:
+              path: /_stcore/health
+              port: http
+            initialDelaySeconds: 10
```
akavel-reply, line 52: do we need such a long wait time? I would expect the container to start much faster, isn't this unnecessarily slowing down our test?

e2e/components/keycloak.yaml
```diff
@@ -0,0 +1,147 @@
...
+            httpGet:
+              path: /health/started
+              port: health
+            periodSeconds: 10
+            failureThreshold: 30
+          readinessProbe:
+            httpGet:
+              path: /health/ready
+              port: health
+            initialDelaySeconds: 20
```
akavel-reply, line 115: is there a reason for this when we already have `startupProbe`?

e2e/components/keycloak.yaml
```diff
@@ -0,0 +1,147 @@
...
+            periodSeconds: 10
+            failureThreshold: 12
+          livenessProbe:
+            httpGet:
+              path: /health/live
+              port: health
+            initialDelaySeconds: 30
```
akavel-reply, line 122: is there a reason for this when we already have `startupProbe`?

e2e/components/litellm.yaml
```diff
@@ -0,0 +1,135 @@
...
+            httpGet:
+              path: /health/liveliness
+              port: http
+            periodSeconds: 10
+            failureThreshold: 30
+          readinessProbe:
+            httpGet:
+              path: /health/readiness
+              port: http
+            initialDelaySeconds: 20
```
akavel-reply, line 103: is there a reason for this when we already have `startupProbe`?

e2e/components/litellm.yaml
```diff
@@ -0,0 +1,135 @@
...
+            periodSeconds: 10
+          livenessProbe:
+            httpGet:
+              path: /health/liveliness
+              port: http
+            initialDelaySeconds: 40
```
akavel-reply, line 109: is there a reason for this when we already have `startupProbe`?

e2e/components/postgres.yaml
```diff
@@ -0,0 +1,99 @@
...
+          volumeMounts:
+            - name: postgres-data
+              mountPath: /var/lib/postgresql/data
+          readinessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - pg_isready -U "$POSTGRES_USER" -d "$POSTGRES_DB"
+            initialDelaySeconds: 10
```
akavel-reply, line 65: can we replace this with fast, frequent `startupProbe` to speed-up the total test time?

e2e/components/postgres.yaml
```diff
@@ -0,0 +1,99 @@
...
+            periodSeconds: 10
+          livenessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - pg_isready -U "$POSTGRES_USER" -d "$POSTGRES_DB"
+            initialDelaySeconds: 30
```
akavel-reply, line 73: can we replace this with fast, frequent `startupProbe` to speed-up the total test time?

e2e/components/chatbot1.yaml
```diff
@@ -0,0 +1,74 @@
...
+# component (components/litellm.yaml) — it's the same LITELLM_MASTER_KEY
+# value, so the plugin's live /v1/models probe against that key succeeds.
+apiVersion: v1
+kind: Secret
+metadata:
+  name: chatbot1-secrets
+  namespace: ${NAMESPACE}
+type: Opaque
+stringData:
+  LITELLM_API_KEY: sk-e2elitellmmasterkey001
```
akavel-reply, line 14: Shouldn't this be a `change-me` value, instead of a real LiteLLM API key?? From what I see, `provision_chatbot1_key.py` seems to be replacing this secret. I am now confused - if we want to test the value from the python script, we should make sure the value before is not correct, otherwise if the script is broken, we might use this one instead. I'm now unsure which one is actually used in the test, and a good quality and trustworthy test should be clear rather than confusing.

e2e/litellm_chatbot_to_catalog_api/README.md
```diff
@@ -0,0 +1,117 @@
...
+Desktop/Engine, and `envsubst` from your OS's `gettext` package). A `kind`
+older than 0.32 will fail with `unknown containerd config version: 4` on
+Docker Desktop.
+
+## Run it locally
+
```
akavel-reply, line 24: This whole following writeup seems duplicated between this file and `.github/workflows/README.md`. We must avoid such duplicadtion, which will quickly and easily degrade into desynchronization. I recommend changing the readme in `.github/workflows/` to just have a reference: "For more information, see: e2e/litellm_chatbot_to_catalog_api/README.md".

After updating this, please make sure to apply review comments from that file into here as well as relevant.

e2e/scripts/create-environment.sh
```diff
@@ -0,0 +1,221 @@
...
+echo "==> Seeding starting dataset"
+apply "${SCENARIO_DIR}/seed/seed-job.yaml"
+kubectl -n "${NAMESPACE}" wait --for=condition=complete "job/seed-${ENV_ID}" --timeout=3m
+
+kubectl -n "${NAMESPACE}" get pods
+
+if [ -n "${GITHUB_ENV:-}" ]; then
+  echo "ENV_ID=${ENV_ID}" >> "${GITHUB_ENV}"
+  echo "NAMESPACE=${NAMESPACE}" >> "${GITHUB_ENV}"
+fi
```
akavel-reply, line 218: Do we need this? it seems ENV_ID and NAMESPACE are both actually the same value as was passed into `--env-id ID` flag, so the caller knows both those values already?

akavel-reply:
I tried running locally.

On first try, catalog didn't start because of catalog configuration changes in https://github.com/naira-project/naira/commit/7228b958fb550e5061c32315252d7f502b23b525.

I reverted that commit and tried running again.

The script failed with the following output:
```console
$ ./e2e/scripts/create-environment.sh --scenario litellm_chatbot_to_catalog_api --env-id local2 --tag localtag2
(...)
==> Seeding starting dataset
==> Applying seed-job.yaml
serviceaccount/seed created
role.rbac.authorization.k8s.io/seed created
rolebinding.rbac.authorization.k8s.io/seed created
job.batch/seed-local2 created
error: timed out waiting for the condition on jobs/seed-local2
```
I think it could be useful to log detailed logs from pods in case of stopping on error (maybe through `trap EXIT`?), so that there'd be more information for debugging.

FWIW, from manual viewing of Pod logs, there's some `urllib.error.HTTPError: HTTP Error 500: Internal Server Error`, seemingly around the following line:
```
│   File "/seed/provision_chatbot1_key.py", line 40, in generate_scoped_key                                                                                                            │
│     with urllib.request.urlopen(req, timeout=10) as resp:                                                                                                                            │
│          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^   
```

.github/workflows/e2e.yml
jagdeepm: Done — replaced the uses now jdx/mise-action, reading the same mise.toml versions local dev uses.

.github/workflows/e2e.yml
jagdeepm: Removed — || true already covers it, no need to also hide stderr

.github/workflows/e2e.yml
jagdeepm: Done

.github/workflows/README.md
jagdeepm: Fixed — this file no longer duplicates the walkthrough at all now , now it just says NS=local directly

e2e/scripts/create-environment.sh
jagdeepm: You're right, they were always the same value — collapsed to just NAMESPACE everywhere the templates use it . ENV_ID stays only as the script's own CLI-facing variable.

e2e/components/chatbot1/app.py
jagdeepm: Removed — chatbot1.yaml always sets this. Now uses os.environ["LITELLM_API_URL"].

e2e/components/chatbot1/app.py
jagdeepm: Same fix applied to all three env vars here (LITELLM_API_URL, LITELLM_API_KEY, LITELLM_MODEL) — all required now, no defaults.

e2e/components/chatbot1.yaml
jagdeepm: Done — replaced the fixed initialDelaySeconds: 10 with a fast (periodSeconds: 2) startupProbe.

e2e/components/keycloak.yaml
jagdeepm: Removed

e2e/components/keycloak.yaml
jagdeepm: Removed

e2e/components/litellm.yaml
jagdeepm: Removed

e2e/components/litellm.yaml
jagdeepm: Removed

e2e/components/postgres.yaml
jagdeepm: Done — added a fast (periodSeconds: 2) startupProbe using the same pg_isready check, dropped the fixed initialDelaySeconds from readiness/liveness.

e2e/components/postgres.yaml
jagdeepm: Same fix, covers both probes.

e2e/components/chatbot1.yaml
jagdeepm: Changed to change-me;

e2e/litellm_chatbot_to_catalog_api/README.md
jagdeepm: Done — .github/workflows/README.md now just points here

e2e/scripts/create-environment.sh
jagdeepm: Removed, always passes --env-id itself now

jagdeepm:
> I tried running locally.
> 
> On first try, catalog didn't start because of catalog configuration changes in [7228b95](https://github.com/naira-project/naira/commit/7228b958fb550e5061c32315252d7f502b23b525).
> 
> I reverted that commit and tried running again.
> 
> The script failed with the following output:
> 
> ```
> $ ./e2e/scripts/create-environment.sh --scenario litellm_chatbot_to_catalog_api --env-id local2 --tag localtag2
> (...)
> ==> Seeding starting dataset
> ==> Applying seed-job.yaml
> serviceaccount/seed created
> role.rbac.authorization.k8s.io/seed created
> rolebinding.rbac.authorization.k8s.io/seed created
> job.batch/seed-local2 created
> error: timed out waiting for the condition on jobs/seed-local2
> ```
> 
> I think it could be useful to log detailed logs from pods in case of stopping on error (maybe through `trap EXIT`?), so that there'd be more information for debugging.
> 
> FWIW, from manual viewing of Pod logs, there's some `urllib.error.HTTPError: HTTP Error 500: Internal Server Error`, seemingly around the following line:
> 
> ```
> │   File "/seed/provision_chatbot1_key.py", line 40, in generate_scoped_key                                                                                                            │
> │     with urllib.request.urlopen(req, timeout=10) as resp:                                                                                                                            │
> │          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^   
> ```

There was changes in base branch that broke the E2E tests and infra so had to refactor some code to make it work now , new successful test workflow run https://github.com/naira-project/workflow-testing/actions/runs/35214875947/job/105180873388

akavel-reply (approved):
Niiiiice, now we're talking! 😃 there are still things that could be improved, but now I'm sincerely OK with merging this at the current state as a step in a good direction! 🥳

As I understand, this workflow won't get executed automatically, but will require a manual action, is that right? It would be awesome if you could showcase how to do this on a tech coffee session

jagdeepm:
> Niiiiice, now we're talking! 😃 there are still things that could be improved, but now I'm sincerely OK with merging this at the current state as a step in a good direction! 🥳
> 
> As I understand, this workflow won't get executed automatically, but will require a manual action, is that right? It would be awesome if you could showcase how to do this on a tech coffee session

Sure We can have a short discussion in a coffee call after this PR is merged , for now we need to trigger manually and pass a PR number

hosseinsalahi (approved):
All open topics are closed. LGTM.

Issue #25, opened 2026-05-28
Software Catalog


mkorbi:
# Software Catalog

## 📋 Overview

**User Story:** As **App Dev Abdel** (Application Developer), I want to browse the software components, repositories, documentation, deployments, and AI dependencies of my application landscape in Naira, so that I can understand what already exists, reuse internal services safely, and see how my application connects to models, datasets, inference endpoints, gateways, and MCP integrations.

**User Story:** As **Platform Peter** (Platform Engineer), I want Naira to automatically discover software components and their runtime relationships from source control, Kubernetes, and GitOps systems, so that I can understand dependency impact before changing platform services such as gateways, inference runtimes, observability stacks, GPU clusters, or shared AI APIs.

**User Story:** As **AI Engineer rAIner**, I want to see which software components consume a model, dataset, inference endpoint, or MCP integration, so that I can understand downstream usage, coordinate changes, and support application teams during model upgrades or deprecations.

The Software Catalog extends Naira's AI asset catalog with the missing software/application context: systems, components, repositories, owners, documentation, deployments, and dependencies. 

---

## 🎯 Context and Problem Statement

Naira's early catalog work focuses on AI assets such as models, datasets, inference endpoints, gateway routes, and integrations. However, platform and AI teams also need to understand the **software components** that consume and expose those assets. Without that context, users can answer questions like "which model exists?" but not reliably answer questions like:

- Which applications are using this model or inference endpoint?
- Which repository owns this service?
- Which team should be contacted before changing a gateway route, model version, or dataset contract?
- Which deployment in Kubernetes corresponds to this software component?
- Which ArgoCD or Flux application manages it?
- Which observability dashboard or logs should I open for this component?
- What downstream applications may be affected by platform changes?

---

### 1. Software Component Entity Model

Naira defines first-class software catalog entity types in the Context Engine. The minimum entity set should include:

- **System** — a higher-level product, application domain, or service group.
- **Component** — a deployable or logical software unit, such as service, frontend, worker, API, job, or AI application.
- **Repository** — the source code repository associated with a component.
- **Deployment / Runtime Workload** — the runtime representation of a component, such as Kubernetes Deployment, StatefulSet, Job, ArgoCD Application, or Flux Kustomization/HelmRelease.
- **Owner / Team** — the person, team, or group responsible for the component.
- **Documentation Link** — human-readable documentation, runbooks, architecture docs, or API docs associated with the component.

[...]

PR #155, opened 2026-08-19
feat(plugins): add GitHub plugin

libera13:
## Description

Adds GitHub integration that connects Kubernetes Deployments to verified GitHub repositories.

Changes
- Add a GitHub plugin that verifies artifact attestations using gh, fetches repository metadata, and extracts CODEOWNERS ownership.
- Link FluxCD GitRepositories to stable GitHub repository identities and support deployment-to-repository relationships.

issue: https://github.com/naira-project/naira/issues/140

## Type of Change

## How Has This Been Tested?
- [x] `go test ./...` passes

## Screenshots
Screenshot 2026-09-15 at 08 51 17


catalog/internal/httpapi/router.go
```diff
@@ -34,7 +38,9 @@ func NewRouter(catalogService *catalog.Service, runner *pluginrun.Runner, logger
 	})
 
 	router.Route("/v1", func(r chi.Router) {
-		r.Use(authMiddleware)
+		if authEnabled {
+			r.Use(authMiddleware)
+		}
```
kocakkaan: I could not understand why we have an option like this. I thought even for local development, we need to have Keycloak enabled so that we would see it from the perspective of the user as much as possible.

plugins/cmd/depl_calls_svc/main.go
```diff
@@ -60,7 +60,7 @@ func (p *Plugin) Collect(ctx context.Context) (pluginapi.CollectResponse, error)
 func (p *Plugin) collect(ctx context.Context, dyn dynamic.Interface) (pluginapi.CollectResponse, error) {
 	var claims pluginapi.CollectResponse
 
-	namespaces, clusterID, err := kubeutil.NamespacesAndClusterID(ctx, dyn)
+	namespaces, clusterID, err := kubeutil.NamespacesAndClusterIDDynamic(ctx, dyn)
```
kocakkaan: Do we have a specific reason for this naming change, would this provide additional information to our side or not?

plugins/cmd/fluxcd/main.go
```diff
@@ -3,7 +3,9 @@
 //
 // TODO: add support for Bucket and other FluxCD sources kinds.
 // TODO: add support for any other resources managed by FluxCD (Services, Ingresses, thirdparty CRDs, ...)
-// TODO: also emit git repository URLs as Nodes (tricky because hostnames & IPs can be local)
+//
+// GitRepository nodes currently support GitHub URLs only. Other Git hosts are
+// skipped until host-specific canonicalization and enrichment are implemented.
```
kocakkaan: I know that this PR resolves this TODO mostly but should we provide another TODO here for the `host-specific canonicalization and enrichment are implemented.` part?

plugins/cmd/github/codeowners.go
```diff
@@ -0,0 +1,47 @@
...
+	out := make([]string, 0, len(in))
+	for _, s := range in {
+		if _, ok := seen[s]; ok {
+			continue
+		}
+		seen[s] = struct{}{}
+		out = append(out, s)
+	}
+	return out
+}
```
kocakkaan, line 47: Maybe we can put this function inside a shared util package?

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,103 @@
...
+		if err := json.NewDecoder(resp.Body).Decode(out); err != nil {
+			return false, fmt.Errorf("decoding github api response for %s: %w", path, err)
+		}
+	}
+	return true, nil
+}
+
+func (c *githubClient) GetRepo(ctx context.Context, owner, repo string) (ghRepo, bool, error) {
+	var r ghRepo
+	found, err := c.do(ctx, fmt.Sprintf("/repos/%s/%s", url.PathEscape(owner), url.PathEscape(repo)), &r)
```
kocakkaan: can we do this something like `githubRepository` instead of `r`?

kocakkaan (commented):
There are just some minor changes/questions from my side. Overall, it looks good.

catalog/internal/httpapi/router.go
libera13: It's true. I had problems with deploying it to the shared cluster. Do you suggest to fixing it there and remove this code?

plugins/cmd/github/codeowners.go
libera13: I would prefer to move that when at least 2 plugins will use it

plugins/cmd/fluxcd/main.go
libera13: Added

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,103 @@
...
+
+func newGithubClient(httpClient *http.Client, baseURL, token string) *githubClient {
+	return &githubClient{
+		httpClient: httpClient,
+		baseURL:    strings.TrimSuffix(baseURL, "/"),
+		token:      token,
+	}
+}
+
+func (c *githubClient) do(ctx context.Context, path string, out any) (bool, error) {
```
kocakkaan: Would there be a better naming for this function? @libera13

kocakkaan (approved):
One small comment, other than that looks good to me 👍

plugins/cmd/github/github.go
libera13: do -> get

kocakkaan:
The functionality provided by this PR must be addressed in #159 .

plugins/cmd/github/github.go
kocakkaan: Thank you 👍

plugins/cmd/fluxcd/main.go
```diff
@@ -3,7 +3,9 @@
 //
 // TODO: add support for Bucket and other FluxCD sources kinds.
 // TODO: add support for any other resources managed by FluxCD (Services, Ingresses, thirdparty CRDs, ...)
-// TODO: also emit git repository URLs as Nodes (tricky because hostnames & IPs can be local)
+//
+// GitRepository nodes currently support GitHub URLs only. Other Git hosts are
+// TODO: add support for other Git hosts.
```
Daviidg: This comment gets cut out

Daviidg (commented):
I see in software catalog we have both github plugin but also things like deployments and services.

Do you think they should be under the same catalog or should they be separated? To me software catalog and things like deployment and services should be separate

Daviidg:
Shouldnt we add the two new plugins to `plugins.yaml`?

plugins/cmd/fluxcd/main.go
```diff
@@ -3,7 +3,9 @@
 //
 // TODO: add support for Bucket and other FluxCD sources kinds.
 // TODO: add support for any other resources managed by FluxCD (Services, Ingresses, thirdparty CRDs, ...)
-// TODO: also emit git repository URLs as Nodes (tricky because hostnames & IPs can be local)
+//
+// GitRepository nodes currently support GitHub URLs only. Other Git hosts are
```
Daviidg: This comment is missing the final part

.github/workflows/dev-publish.yml
```diff
@@ -120,13 +120,28 @@ jobs:
       platforms: "linux/amd64,linux/arm64"
     secrets:
       registry-password: ${{ secrets.GITHUB_TOKEN }}
-
   container-publish-plugin-mcp-servers:
     name: Publish · Plugin MCP Servers
     uses: naira-project/naira-github-workflows/.github/workflows/reusable-container-build.yml@de9bdb0d30bec81f6730a82d28f7b7acf481609f
     with:
       image-name: "naira-plugin-mcp-servers"
       dockerfile: "plugins/cmd/mcp_servers/Dockerfile"
```
Daviidg, line 129: I think some of the fields of the `container-publish-plugin-mcp-servers` got cut out when you merged. It should have `context`, `platform` & `secret` fields like the others

plugins/cmd/depl_from_repo/main.go
```diff
@@ -0,0 +1,149 @@
...
+
+func New(cfg config, logger *log.Logger) (*Plugin, error) {
+	return &Plugin{config: cfg, logger: logger}, nil
+}
+
+func main() {
+	app := pluginmain.New[config]()
+
+	p, err := New(app.PluginConfig, app.Logger)
+	if err != nil {
```
akavel-reply: the `New` function seems to always return nil error, so this looks like dead code (and the err return seems unnecessary first of all)

.github/workflows/dev-publish.yml
```diff
@@ -120,7 +120,6 @@ jobs:
       platforms: "linux/amd64,linux/arm64"
     secrets:
       registry-password: ${{ secrets.GITHUB_TOKEN }}
-
```
akavel-reply, line 123: why remove empty line here? other items above seem to have empty line separators

similar question about the new block below

plugins/internal/repositorydiscovery/discovery.go
```diff
@@ -0,0 +1,229 @@
...
+	if strings.Contains(value, "://") || strings.Contains(value, "@") {
+		return "", "", false
+	}
+	value = strings.TrimSuffix(value, ".git")
+	parts := strings.Split(strings.Trim(value, "/"), "/")
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", false
+	}
+	return parts[0], parts[1], true
+}
```
akavel-reply: As we discussed, this function's body is full of holes in many ways. Also code around it needs to be improved for better quality, more thoughtfully written

plugins/internal/repositorydiscovery/discovery.go
```diff
@@ -0,0 +1,229 @@
...
+	ClusterID  string
+	Namespace  string
+	Deployment string
+	Images     []string
+	Repository Repository
+}
+
+var sourceLabelKeys = []string{
+	"org.opencontainers.image.source",
+}
```
akavel-reply: why is it an array if it only has one value

akavel-reply (changes requested):

plugins/internal/repositorydiscovery/discovery.go
libera13: fixed

plugins/cmd/fluxcd/main.go
```diff
@@ -3,7 +3,9 @@
 //
 // TODO: add support for Bucket and other FluxCD sources kinds.
 // TODO: add support for any other resources managed by FluxCD (Services, Ingresses, thirdparty CRDs, ...)
-// TODO: also emit git repository URLs as Nodes (tricky because hostnames & IPs can be local)
+//
+// GitRepository nodes currently support GitHub URLs only. Other Git hosts are ignored.
+// TODO: add support for other Git hosts.
```
akavel-reply: I think the annotation from the old comment is still relevant?

```suggestion
// TODO: add support for other Git hosts (tricky because hostnames & IPs can be local)
```

plugins/cmd/fluxcd/main.go
```diff
@@ -104,11 +107,17 @@ func (p *Plugin) collect(ctx context.Context, disc discovery.DiscoveryInterface,
 	repoByPath := map[string]pluginapi.NodeID{} // "ns/name" → NodeID
 	for _, r := range repos {
 		shortPath := r.GetNamespace() + "/" + r.GetName()
+		url, _, _ := unstructured.NestedString(r.Object, "spec", "url")
+		repoPath := repositoryidentity.GitHubRepositoryNodePathFromURL(url)
+		if repoPath == "" {
+			log.Printf("%s: WARN: skipping GitRepository %s with unsupported URL %q", pluginName, shortPath, url)
+			continue
```
akavel-reply: why remove functionality from the plugin? that's very surprising to me; is it on purpose? I don't see it clearly explained and rationalized anywhere obvious

plugins/cmd/fluxcd/main.go
```diff
@@ -104,11 +107,17 @@ func (p *Plugin) collect(ctx context.Context, disc discovery.DiscoveryInterface,
...
+		}
+
 		id := pluginapi.NodeID{
 			Kind: pluginapi.NodeKindGitRepository,
-			Path: clusterID + "/" + shortPath,
+			Path: repoPath,
```
akavel-reply: huh? again, this is a major change, and I don't see it discussed anywhere, with pros & cons & rationale

plugins/cmd/fluxcd/main_test.go
```diff
@@ -78,13 +78,13 @@ func TestCollect(t *testing.T) {
 			want: pluginapi.CollectResponse{
 				Nodes: []pluginapi.NodeClaim{
 					{ID: nodeID("Kustomization.fluxcd", "flux-system/my-app")},
-					{ID: nodeID("git_repository", "flux-system/my-repo"),
+					{ID: githubRepositoryNodeID("example", "repo"),
```
akavel-reply: please bring back the explicit `nodeID` calls, with string kind.

plugins/cmd/github/attestation.go
```diff
@@ -0,0 +1,116 @@
...
+type ghAttestationEntry struct {
+	VerificationResult struct {
+		Signature struct {
+			Certificate ghCertificate `json:"certificate"`
+		} `json:"signature"`
+	} `json:"verificationResult"`
+}
+
+// signature.certificate is populated directly from the OpenID Connect token that GitHub has generated
+// so it's properties contain values that cannot be manipulated by the workflow that originated the attestation.
```
akavel-reply: typo/grammar

```suggestion
// so its properties contain values that cannot be manipulated by the workflow that originated the attestation.
```

plugins/cmd/github/attestation.go
```diff
@@ -0,0 +1,116 @@
...
+// `gh attestation verify --format json` output
+type ghAttestationEntry struct {
+	VerificationResult struct {
+		Signature struct {
+			Certificate ghCertificate `json:"certificate"`
+		} `json:"signature"`
+	} `json:"verificationResult"`
+}
+
+// signature.certificate is populated directly from the OpenID Connect token that GitHub has generated
```
akavel-reply: I don't understand this whole comment. What `signature.certificate`? it's also referring to some manipulation by some workflow - huh? I don't understand anything about what it's trying to tell me, or why is it relevant to me here.

plugins/cmd/github/attestation.go
```diff
@@ -0,0 +1,116 @@
...
+		ownerFromCert, nameFromCert, ok := repositoryidentity.ParseGitHubRepository(cert.SourceRepositoryURI)
+		if !ok {
+			continue
+		}
+		if !strings.EqualFold(ownerFromCert, org) {
+			continue
+		}
+		return ownerFromCert, nameFromCert, true, nil
+	}
+	return "", "", false, nil
```
akavel-reply: We could reduce the number of returned values from this func (remove the boolean) if we returned a predefined sentinel error here (e.g. `var ErrAttestationMissing = errors.New("attestation not found for given image")` or something similar - see e.g.: https://pkg.go.dev/io#pkg-variables).

plugins/internal/repositoryidentity/github.go
```diff
@@ -0,0 +1,83 @@
+package repositoryidentity
+
+import (
+	"net/url"
+	"regexp"
+	"strings"
+)
+
+var (
+	// Format SCP: git@github.com:owner/repo[.git]
```
akavel-reply: what does this "SCP" mean? where is it defined, do you have some online reference you could add here?

plugins/internal/repositoryidentity/github.go
```diff
@@ -0,0 +1,83 @@
...
+	if len(parts) != 2 {
+		return "", "", false
+	}
+
+	return validate(parts[0], parts[1])
+}
+
+func validate(rawOwner, rawRepo string) (owner, name string, ok bool) {
+	repo := strings.TrimSuffix(rawRepo, ".git")
+	if repo == "." || repo == ".." {
```
akavel-reply: couldn't we have a stricter check here? I feel pretty sure github org & account names have much more limited set of characters allowed, and probably also the first character is even more limited.

plugins/internal/repositoryidentity/github.go
```diff
@@ -0,0 +1,83 @@
...
+		return "", "", false
+	}
+	return rawOwner, repo, true
+}
+
+// GitHubRepositoryNodePath returns the stable graph path for a GitHub repository.
+func GitHubRepositoryNodePath(owner, name string) string {
+	if owner == "" || name == "" {
+		return ""
+	}
```
akavel-reply: This is a bit weird to me: why we only check those two cases? what does it give us if we return an empty string in those, but don't return an empty string in other invalid cases? should we run `validate` here and return error if invalid?

plugins/internal/repositoryidentity/github.go
```diff
@@ -0,0 +1,83 @@
...
+
+func validate(rawOwner, rawRepo string) (owner, name string, ok bool) {
+	repo := strings.TrimSuffix(rawRepo, ".git")
+	if repo == "." || repo == ".." {
+		return "", "", false
+	}
+	return rawOwner, repo, true
+}
+
+// GitHubRepositoryNodePath returns the stable graph path for a GitHub repository.
```
akavel-reply:
```suggestion
// GitHubRepositoryNodePath returns a stable node path for a GitHub repository.
```

plugins/internal/repositoryidentity/github.go
```diff
@@ -0,0 +1,83 @@
...
+func GitHubRepositoryNodePath(owner, name string) string {
+	if owner == "" || name == "" {
+		return ""
+	}
+	return "github.com/" + strings.ToLower(owner+"/"+name)
+}
+
+// GitHubRepositoryNodePathFromURL returns the stable graph path for a GitHub
```
akavel-reply:
```suggestion
// GitHubRepositoryNodePathFromURL returns a stable node path for a GitHub
```

plugins/cmd/github/attestation.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (v *attestationVerifier) CheckGhAvailable(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, v.timeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, v.ghPath, "--version")
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
```
akavel-reply: would it make sense to run cmd.CombinedOutput() instead of collecting only stderr?

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,271 @@
...
+	return entry.Images[0], true
+}
+
+// imageReferencesOrg is a filter to skip the network calls for images that are not in
+// the configured org. It is not a security check - it's an optimization
+func imageReferencesOrg(image, org string) bool {
+	if org == "" {
+		return false
+	}
+	return strings.Contains(strings.ToLower(image), strings.ToLower(org))
```
akavel-reply: `Contains` here seems wrong

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,271 @@
...
+	if len(entry.Images) != 1 {
+		return "", false
+	}
+	return entry.Images[0], true
+}
+
+// imageReferencesOrg is a filter to skip the network calls for images that are not in
+// the configured org. It is not a security check - it's an optimization
+func imageReferencesOrg(image, org string) bool {
+	if org == "" {
```
akavel-reply: wasn't this mandatory?

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,271 @@
...
+		if err != nil {
+			p.logger.Printf("verifying attestation for %s (deployment %s/%s): %v", image, entry.Namespace, entry.Name, err)
+			continue
+		}
+		if !verified {
+			p.logger.Printf("attestation for %s (deployment %s/%s) not verified for org %q", image, entry.Namespace, entry.Name, p.config.GitHubOrg)
+			continue
+		}
+
+		repoNodeID := gitRepositoryNodeID(owner, name)
```
akavel-reply: I think we could:

1. introduce `type ownerAndName struct { owner, name string }`
2. remove `gitRepositoryNodeID` function
3. remove the `repoCache` type and just use a `map[ownerAndName]bool` instead.

`collectRepo()` could then take `ownerAndName` as argument.

`ownerAndName` could also get a `.ToNodeID()` method for use when building NodeClaim/RelationClaim values.

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,104 @@
...
+	httpClient *http.Client
+	baseURL    string // e.g. https://api.github.com
+	token      string
+}
+
+type ghRepo struct {
+	HTMLURL  string `json:"html_url"`
+	Language string `json:"language"`
+	Homepage string `json:"homepage"`
+}
```
akavel-reply, line 31: Can you please provide a link to documentation of those fields in the official GitHub API docs? I'm having trouble finding them, and notably I'm confused what's the difference between HTMLURL and Homepage. Also, why do we need Language?

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,104 @@
...
+	return true, nil
+}
+
+func (c *githubClient) GetRepo(ctx context.Context, owner, repo string) (ghRepo, bool, error) {
+	var githubRepo ghRepo
+	found, err := c.get(ctx, fmt.Sprintf("/repos/%s/%s", url.PathEscape(owner), url.PathEscape(repo)), &githubRepo)
+	if err != nil {
+		return ghRepo{}, false, fmt.Errorf("getting repo %s/%s: %w", owner, repo, err)
+	}
+	return githubRepo, found, nil
```
akavel-reply: we could remove the boolean return value if we returned a predefined known error value here

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return false, fmt.Errorf("calling github api %s: %w", path, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return false, nil
```
akavel-reply: we could return a known predefined error value here instead and remove the boolean return value completely

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+type ghRepo struct {
+	HTMLURL  string `json:"html_url"`
+	Language string `json:"language"`
+	Homepage string `json:"homepage"`
+}
+
+type ghContent struct {
+	Content  string `json:"content"` // base64, when Encoding == "base64"
```
akavel-reply: the comment here is confusing; also, do you have reference link for this on the internet?

akavel-reply (commented):
First bunch of comments.

plugins/cmd/fluxcd/main.go
libera13: Before my changes, FluxCD used the Flux GitRepository CRD ID as the nodeID.path. I changed that so it uses the parsed Github repository URL as the nodeID.path.

I tested this with GitHub, so I know it works.

In flux plugin I see the skipping of:
a) Unrecognized URLs in GitHub
b) URLs from hosts other than GitHub

To bring back the previous functionality of FluxCD, as a fallback I could:
- Use the previous Flux GitRepository CRD ID as the nodeID.path. This will work, but for me, it would be misleading to have two different ways of constructing the nodeID.path.
- Use the Git repo URL. Here I could either try to parse it or not, but either way, it’s something I haven't tested
- Keep the previous code in Flux, but there will be two nodes from now on in flux connected with gitrepos. The previous node kind git_repository will use the kind GitRepository.fluxcd. The node where the path is a GitHub URL will use the git_repository node kind.

plugins/cmd/github/attestation.go
libera13: ~~If I use cmd.CombinedOutput(), both stdout and stderr will end up in a single buffer, which isn't what I want. If the command succeeds and outputs JSON to stdout, but also prints a warning to stderr, the JSON parsing below will fail.~~

~~or maybe you mean to log stdout in case of failure too?~~

plugins/internal/repositoryidentity/github.go
libera13: We could, but I didn't want to do that. If they change anything about the allowed characters, this code will be outdated then

plugins/internal/repositoryidentity/github.go
libera13: I changed that to:
```
// GitHubRepositoryNodePath returns a stable node path for a GitHub repository.
// owner and name must be non-empty and must not contain "/" — e.g. as
// returned by ParseGitHubRepository.
func GitHubRepositoryNodePath(owner, name string) string {
	return "github.com/" + strings.ToLower(owner+"/"+name)
}
```
In the current code, in all cases where this func is used, the owner and name are already taken from ParseGitHubRepository

plugins/cmd/github/github.go
libera13: HTMLURL is link to github repo.
Homepage is project website, like https://naira-project.github.io/

https://github.com/naira-project/naira/issues/25 is provided this list: "Git source control: repository metadata, default branch, description, topics/tags, ownership hints, README/documentation links, CODEOWNERS where available."
I wanted to have minimal amount of props for now, that's why I included only few repository metadata like Language, Homepage and HTMLURL

plugins/internal/repositoryidentity/github.go
libera13: After discussion I will add this regexp to be even more defensive here.
Worth to remember naming convention can change in future, as written in some github issue "naming isn't an ongoing limit" https://github.com/github/docs/issues/44518#issuecomment-4674516090

plugins/cmd/fluxcd/main.go
libera13: I implemented the last point "Keep the previous code in Flux..."

plugins/cmd/github/attestation.go
libera13: I rewrite the comment, please take a look now if it's more clear

plugins/internal/repositoryidentity/github.go
libera13: added link

plugins/cmd/github/main.go
libera13: I changed that to filter out ghcr/{other-orgs} images. Images from other registries pass through this filter untouched

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,260 @@
...
+	imageLower := strings.ToLower(image)
+
+	// Fast-path for ghcr.io based on the ghcr.io/<org>/<repo> naming convention.
+	if strings.HasPrefix(imageLower, "ghcr.io/") {
+		orgLower := strings.ToLower(p.config.GitHubOrg)
+		ghcrPrefix := fmt.Sprintf("ghcr.io/%s/", orgLower)
+		return strings.HasPrefix(imageLower, ghcrPrefix)
+	}
+
+	return true
```
akavel-reply: Please refactor to "early return" pattern (also then, what do you mean by fast-path?)

```go
if !strings.HasPrefix(imageLower, "ghcr.io/") {
    return true
}

orgLower := ...
```

plugins/cmd/github/attestation.go
```diff
@@ -0,0 +1,119 @@
...
+// in the organization and, if valid, returns the repository owner and name.
+func (v *attestationVerifier) Verify(ctx context.Context, image, org string) (ownerAndName, error) {
+	ctx, cancel := context.WithTimeout(ctx, v.timeout)
+	defer cancel()
+
+	// passing gh an oci:// reference (rather than a resolved digest) means
+	// gh itself talks to the registry to resolve the digest and fetch
+	// the attestation bundle
+	cmd := exec.CommandContext(ctx, v.ghPath, "attestation", "verify",
+		"oci://"+image,
```
akavel-reply: I'm somewhat nervous of passing user-controlled inputs to a CLI subcommand. Fortunately this I believe doesn't go through bash (or other shell), so I don't actually really see how someone could abuse this. Still, could we, as at least a sanity check, try to parse the resulting `oci://...` string as an URL and see if it at least validates in this aspect?

plugins/cmd/github/attestation.go
```diff
@@ -0,0 +1,119 @@
...
+	VerificationResult struct {
+		Signature struct {
+			Certificate ghCertificate `json:"certificate"`
+		} `json:"signature"`
+	} `json:"verificationResult"`
+}
+
+type ghCertificate struct {
+	SourceRepositoryURI      string `json:"sourceRepositoryURI"`
+	SourceRepositoryOwnerURI string `json:"sourceRepositoryOwnerURI"`
```
akavel-reply: Do we need this one? I don't see it used in code?

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,119 @@
...
+	if out != nil {
+		if err := json.NewDecoder(resp.Body).Decode(out); err != nil {
+			return fmt.Errorf("decoding github api response for %s: %w", path, err)
+		}
+	}
+	return nil
+}
+
+func (c *githubClient) GetRepo(ctx context.Context, owner, repo string) (ghRepo, error) {
+	var githubRepo ghRepo
```
akavel-reply:
```suggestion
	var r ghRepo
```

plugins/cmd/github/github.go
```diff
@@ -0,0 +1,119 @@
...
+// GetCodeowners tries the well-known CODEOWNERS locations, in the order
+// GitHub itself checks them, and returns the content of the first one found.
+// Order of locations: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location
+// This function doesn't support files bigger than 1MB, as it would require handling a different encoding than base64.
+func (c *githubClient) GetCodeowners(ctx context.Context, owner, repo string) (string, error) {
+	candidates := []string{".github/CODEOWNERS", "CODEOWNERS", "docs/CODEOWNERS"}
+
+	for _, path := range candidates {
+		var content ghContent
+		err := c.get(ctx, fmt.Sprintf("/repos/%s/%s/contents/%s", url.PathEscape(owner), url.PathEscape(repo), path), &content)
```
akavel-reply: Please add TODO somewhere to avoid downloading files larger than 1MB at all - or otherwise have some limit on the size of data we might read into RAM.

plugins/cmd/github/codeowners.go
```diff
@@ -0,0 +1,47 @@
...
+			continue
+		}
+
+		defaultOwners = append(defaultOwners, fields[1:]...)
+	}
+
+	return dedupeStrings(defaultOwners)
+}
+
+func dedupeStrings(in []string) []string {
```
akavel-reply: maybe we could run `slices.Sort` and `slices.Compact`? This would give us a sorted list as a result, maybe that would be preferable?

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,260 @@
...
+
+	codeowners, err := p.github.GetCodeowners(ctx, repo.owner, repo.name)
+	if errors.Is(err, errGithubResourceNotFound) {
+		return nodes, nil, nil
+	}
+	if err != nil {
+		return nil, nil, fmt.Errorf("fetching codeowners: %w", err)
+	}
+
+	ownerNodes, ownedByRelations := codeownersClaims(repoNodeID, extractDefaultCodeowners(codeowners))
```
akavel-reply: Why we only extract the default codeowners? Please add explanation in a comment in the readme if we want to keep doing it like this, it's not really obvious to me.

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,260 @@
...
+	nodes = append(nodes, ownerNodes...)
+	return nodes, ownedByRelations, nil
+}
+
+func codeownersClaims(repoNodeID pluginapi.NodeID, handles []string) ([]pluginapi.NodeClaim, []pluginapi.RelationClaim) {
+	nodes := make([]pluginapi.NodeClaim, 0, len(handles))
+	relations := make([]pluginapi.RelationClaim, 0, len(handles))
+
+	for _, handle := range handles {
+		ownerNodeID := pluginapi.NodeID{Kind: pluginapi.NodeKindOwner, Path: handle}
```
akavel-reply: Path should be namespaced - in this case, owner is a github user, the same nick could be used by somebody else somewhere else.

Also, please split this to multiline struct definition for better readability and visibility.

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,260 @@
...
+		nodes = append(nodes, pluginapi.NodeClaim{ID: ownerNodeID})
```
akavel-reply: I think this will create duplicate nodes if multiple repos share the same owner, no?

Please make sure to add a regression test for this case.

plugins/cmd/github/main.go
```diff
@@ -0,0 +1,260 @@
...
+			}
+			resp.Nodes = append(resp.Nodes, nodes...)
+			resp.Relations = append(resp.Relations, relations...)
+			repos[repo] = true
+		}
+
+		deploymentNodeClaim := pluginapi.NodeClaim{ID: entry.NodeID()}
+		deploymentRepoRelation := pluginapi.RelationClaim{
+			Kind: pluginapi.RelationKindBuiltFrom,
+			From: entry.NodeID(),
```
akavel-reply: Please rename this variable from `entry` to `deployment`, and the corresponding slice from `entries` to `deployments`. Reading this I wasn't sure what "entry" means, I had do scroll up and jump a few times. I think even the type is actually named `Deployment`, so naming the slice/var same would be nice.

plugins/cmd/fluxcd/main_test.go
```diff
@@ -75,18 +75,20 @@ func TestCollect(t *testing.T) {
...
-						Properties: pluginapi.PropertyMap{"url": "https://github.com/example/repo"}},
-				},
-				Relations: []pluginapi.RelationClaim{
-					{Kind: "sourced_from",
-						From: nodeID("Kustomization.fluxcd", "flux-system/my-app"),
-						To:   nodeID("git_repository", "flux-system/my-repo")},
-				},
-			},
+			want: func() pluginapi.CollectResponse {
+				fluxRepoNode, gitRepoNode, refRel := githubRepoFixture("flux-system", "my-repo", "example", "repo")
```
akavel-reply: Instead of introducing this function and its related complexity, could we:

1. Have one test case with `github.com/...` url, and explicitly listed all relevant `GitRepository.fluxcd` nodes/relations, to easily see the full picture of what is produced for a github repo;
2. For the remaining testcases, use a non-github url, and thus presumably get effectively the same as in original `main_test.go` (just with a change of kind to `"GitRepository.fluxcd"`)?

plugins/cmd/fluxcd/main_test.go
```diff
@@ -324,6 +348,78 @@ func TestCollect(t *testing.T) {
 	}
 }
 
+func TestCollect_ExternalGitRepositoryNode(t *testing.T) {
```
akavel-reply: why is this in a separate new test function? the loop looks practically the same to me at first glance, so I'm confused

plugins/cmd/github/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+FROM alpine:3.20 AS gh-cli
+
+ARG GH_CLI_VERSION=2.63.2
+ARG TARGETARCH=amd64
+
+WORKDIR /gh
+# Download GH CLI and verify file integrity against the downloaded checksums
+RUN apk add --no-cache curl \
+ && curl -sSLO https://github.com/cli/cli/releases/download/v${GH_CLI_VERSION}/gh_${GH_CLI_VERSION}_linux_${TARGETARCH}.tar.gz \
+ && curl -sSLO https://github.com/cli/cli/releases/download/v${GH_CLI_VERSION}/gh_${GH_CLI_VERSION}_checksums.txt \
```
akavel-reply: I never really understand the benefit of downloading and checking checksums from the same host as the checksummed binary - wouldn't the same hacker be able to subvert them both if they got access to one of them? Instead, could it make more sense to hardcode the checksum in an ARG of the dockerfile?

plugins/cmd/github/github.go
libera13: I put limit and added tests to verify it

plugins/cmd/github/codeowners.go
libera13: Hard to choose. Let's preserve order from the CODEOWNERS file maybe it will make it easier to read

plugins/cmd/fluxcd/main_test.go
libera13: I removed it

plugins/cmd/fluxcd/main_test.go
```diff
@@ -227,7 +215,44 @@ func TestCollect(t *testing.T) {
...
+				deployment("team-a", "app",
+					kustLabel("flux-system", "my-app")),
+			},
+			want: pluginapi.CollectResponse{
+				Nodes: []pluginapi.NodeClaim{
+					{ID: nodeID("GitRepository.fluxcd", "flux-system/my-repo"),
+						Properties: pluginapi.PropertyMap{"url": "https://github.com/example/repo"}},
+					{ID: nodeID("Kustomization.fluxcd", "flux-system/my-app")},
+					{ID: nodeID("deployment", "team-a/app")},
+					{ID: externalRepoID("example", "repo"),
```
akavel-reply: As before: please use just raw `nodeID` and remove the `externalRepoID` helper.

plugins/cmd/github/attestation_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// fakeGh creates a tiny gh replacement.
+func fakeGh(t *testing.T, stdout string, exitCode int) string {
+	t.Helper()
+
+	path := filepath.Join(t.TempDir(), "gh")
+	script := fmt.Sprintf("#!/bin/sh\nprintf '%%s' %q\nexit %d\n", stdout, exitCode)
```
akavel-reply: Can we use multiline string here? script will be easier to read (and shell is especially tricky and prone to footguns through tiny changes)

plugins/cmd/github/attestation_test.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ghPath := tt.ghPath
+			if ghPath == "" {
+				ghPath = fakeGh(t, tt.output, tt.exitCode)
+			}
+			verifier := newAttestationVerifier(ghPath, "token", 5*time.Second)
+
+			repo, err := verifier.Verify(
+				context.Background(),
```
akavel-reply: `t.Context()` ?

plugins/cmd/github/attestation_test.go
```diff
@@ -0,0 +1,110 @@
...
+		{
+			name:    "returns an error for malformed JSON",
+			output:  "not-json",
+			org:     "naira-project",
+			wantErr: "parsing gh attestation verify output",
+		},
+		{
+			name:    "returns an error when gh cannot be started",
+			ghPath:  filepath.Join("/tmp", "missing-gh"),
+			org:     "naira-project",
```
akavel-reply: this field seems always having the same value in all testcases, so presumably we can remove it?

plugins/cmd/github/attestation_test.go
```diff
@@ -0,0 +1,110 @@
...
+		{
+			name:      "matches organization case insensitively",
+			output:    attestationJSON("https://github.com/Naira-Project/service"),
+			org:       "naira-project",
+			wantOwner: "Naira-Project",
+			wantName:  "service",
+		},
+		{
+			name:    "rejects a different organization",
+			output:  attestationJSON("https://github.com/other-org/service"),
```
akavel-reply: maybe let's also add a testcases with:
1. a non-github URL (in theory gh shouldn't ever return one, but let's see how we behave)
2. `attestationJSON("")` (valid JSON but empty URL from gh)

plugins/cmd/github/github_test.go
```diff
@@ -0,0 +1,149 @@
...
+	defer srv.Close()
+
+	c := newGithubClient(srv.Client(), srv.URL, "secret-token")
+	repo, err := c.GetRepo(context.Background(), "acme", "service")
+	require.NoError(t, err)
+	assert.Equal(t, "Go", repo.Language)
+}
+
+func TestGithubClient_GetCodeowners(t *testing.T) {
+	const codeownersBody = "* @acme/team\n"
```
akavel-reply: do we need the trailing `\n`?

plugins/cmd/github/github_test.go
```diff
@@ -0,0 +1,149 @@
...
+		{
+			name:         "not found anywhere",
+			wantErr:      errGithubResourceNotFound,
+			wantRequests: []string{".github/CODEOWNERS", "CODEOWNERS", "docs/CODEOWNERS"},
+		},
+		{
+			name: "skips unsupported encoding and keeps looking",
+			responses: map[string]ghContent{
+				"docs/CODEOWNERS": {Content: codeownersBody, Encoding: "text"},
+			},
```
akavel-reply: this doesn't really match the "keeps looking" claim of the test name - where does it "keep looking" in it?

plugins/cmd/github/github_test.go
```diff
@@ -0,0 +1,149 @@
...
+			wantErr:      errGithubResourceNotFound,
+			wantRequests: []string{".github/CODEOWNERS", "CODEOWNERS", "docs/CODEOWNERS"},
+		},
+	}
```
akavel-reply: with only three paths being scanned, would be nice to have a testcase for each of the three ones

plugins/cmd/github/github_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var requests []string
+			srv := newFakeGithubContentsServer(t, "acme", "service", tt.responses, &requests)
+
+			got, err := newGithubClient(srv.Client(), srv.URL, "").GetCodeowners(context.Background(), "acme", "service")
```
akavel-reply: please split for easier reading:

```suggestion
			got, err := newGithubClient(srv.Client(), srv.URL, "").
				GetCodeowners(context.Background(), "acme", "service")
```

plugins/cmd/github/github_test.go
```diff
@@ -0,0 +1,149 @@
...
+			return
+		}
+		_ = json.NewEncoder(w).Encode(content)
+	}))
+	t.Cleanup(srv.Close)
+	return srv
+}
+
+func TestGithubClient_GetCodeownersDoesNotReadOversizedResponse(t *testing.T) {
+	oversizedContentSize := maxGithubResponseBytes + (1 << 20) // + 1MB
```
akavel-reply: for easier reading:
```suggestion
	oversizedContentSize := maxGithubResponseBytes + 1024*1024 // + 1MB
```

plugins/cmd/github/github_test.go
```diff
@@ -0,0 +1,149 @@
...
+				Body:       &countingReadCloser{Reader: bytes.NewReader(responseBody), bytesRead: &bytesRead},
+				Request:    req,
+			}, nil
+		}),
+	}
+
+	_, err := newGithubClient(httpClient, "https://api.github.test", "").GetCodeowners(context.Background(), "acme", "service")
+
+	require.Error(t, err)
+	assert.Greater(t, len(responseBody), 3<<20)
```
akavel-reply: this shouldn't be here but rather just below `responseBody` definition, if at all anywhere. It's not a test assertion, but rather a sanity-check of the test input, and thus if at all anywhere, it should be also a `require` not an `assert`.

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+`
+	path := filepath.Join(t.TempDir(), "gh")
+	require.NoError(t, os.WriteFile(path, []byte(script), 0o755))
+	return path
+}
+
+func newFakeGithubAPI(t *testing.T) *httptest.Server {
+	t.Helper()
+
+	codeowners := base64.StdEncoding.EncodeToString([]byte("* @naira-project/team\n"))
```
akavel-reply: do we need the trailing `\n`?

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+	}))
+	t.Cleanup(srv.Close)
+	return srv
+}
+
+func TestPlugin_Collect(t *testing.T) {
+	const (
+		clusterUID = "test-cluster-uid"
+		namespace  = "default"
```
akavel-reply: can we inline those consts? they don't seem used in many places really, and I feel it'd make the test a bit easier to follow

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+			fmt.Fprintf(w, `{"content":%q,"encoding":"base64"}`, codeowners)
+		default:
+			w.WriteHeader(http.StatusNotFound)
+		}
+	}))
+	t.Cleanup(srv.Close)
+	return srv
+}
+
+func TestPlugin_Collect(t *testing.T) {
```
akavel-reply: I find the contents of this test really hard to read and understand :/ I'd be grateful if you could try to make it easier somehow; some initial suggestions follow, though not sure they'll be enough

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+			name:        "org match, attestation verified: repo, owner and deployment linked",
+			deployments: map[string][]string{"app": {"ghcr.io/naira-project/service-a:v1"}},
+			wantNodes: []pluginapi.NodeClaim{
+				{ID: repoA, Properties: pluginapi.PropertyMap{"url": "https://github.com/naira-project/service-a", "language": "Go"}},
+				{ID: ownerTeam},
+				{ID: deploymentNode("app")},
+			},
+			wantRelations: []pluginapi.RelationClaim{
+				{Kind: pluginapi.RelationKindOwnedBy, From: repoA, To: ownerTeam},
+				{Kind: pluginapi.RelationKindBuiltFrom, From: deploymentNode("app"), To: repoA},
```
akavel-reply: please split for easier reading:

```suggestion
				{Kind: pluginapi.RelationKindOwnedBy,
					From: repoA,
					To: ownerTeam},
				{Kind: pluginapi.RelationKindBuiltFrom,
					From: deploymentNode("app"),
					To: repoA},
```

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+		clusterUID = "test-cluster-uid"
+		namespace  = "default"
+	)
+
+	githubAPI := newFakeGithubAPI(t)
+	ghPath := newFakeGhCLI(t)
+
+	repoA := pluginapi.NodeID{Kind: pluginapi.NodeKindGitRepository, Path: "github.com/naira-project/service-a"}
+	repoE := pluginapi.NodeID{Kind: pluginapi.NodeKindGitRepository, Path: "github.com/naira-project/service-e"}
+	ownerTeam := pluginapi.NodeID{Kind: pluginapi.NodeKindOwner, Path: "github.com/@naira-project/team"}
```
akavel-reply:
```suggestion
	var (
		repoA     = nodeID("git_repository", "github.com/naira-project/service-a")
		repoE     = nodeID("git_repository", "github.com/naira-project/service-e")
		ownerTeam = nodeID("owner", "github.com/@naira-project/team")
	)
```

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+	deploymentNode := func(name string) pluginapi.NodeID {
```
akavel-reply: naming of this func is inconsistent with namings of vars above - it has "Node" suffix (though also it's not a Node but a NodeID), they don't have "Node" suffix

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+		name          string
+		deployments   map[string][]string // deployment name -> container images
+		wantNodes     []pluginapi.NodeClaim
+		wantRelations []pluginapi.RelationClaim
+	}{
+		{
+			name:        "org match, attestation verified: repo, owner and deployment linked",
+			deployments: map[string][]string{"app": {"ghcr.io/naira-project/service-a:v1"}},
+			wantNodes: []pluginapi.NodeClaim{
+				{ID: repoA, Properties: pluginapi.PropertyMap{"url": "https://github.com/naira-project/service-a", "language": "Go"}},
```
akavel-reply: please split to multiline

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+				{ID: repoE, Properties: pluginapi.PropertyMap{"url": "https://github.com/naira-project/service-e", "language": "Go"}},
+				{ID: ownerTeam}, // only once, even though both repos share it
+				{ID: deploymentNode("app-a")},
+				{ID: deploymentNode("app-e")},
+			},
+			wantRelations: []pluginapi.RelationClaim{
+				{Kind: pluginapi.RelationKindOwnedBy, From: repoA, To: ownerTeam},
+				{Kind: pluginapi.RelationKindOwnedBy, From: repoE, To: ownerTeam},
+				{Kind: pluginapi.RelationKindBuiltFrom, From: deploymentNode("app-a"), To: repoA},
+				{Kind: pluginapi.RelationKindBuiltFrom, From: deploymentNode("app-e"), To: repoE},
```
akavel-reply: Please use string Kinds and split the structs to multiline (see comments above)

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+	repoE := pluginapi.NodeID{Kind: pluginapi.NodeKindGitRepository, Path: "github.com/naira-project/service-e"}
+	ownerTeam := pluginapi.NodeID{Kind: pluginapi.NodeKindOwner, Path: "github.com/@naira-project/team"}
+
+	deploymentNode := func(name string) pluginapi.NodeID {
+		return pluginapi.NodeID{Kind: pluginapi.NodeKindDeployment, Path: clusterUID + "/" + namespace + "/" + name}
+	}
+
+	tests := []struct {
+		name          string
+		deployments   map[string][]string // deployment name -> container images
```
akavel-reply: 1. please include mention of images in name, otherwise reading the following tests becomes very confusing
2. all testcases seem to have only one image, so use and handling of `[]string` here is confusing and seemingly unnecessary - I recommend either adding test with multiple images or simplifying this type

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+func TestSingleContainerImage(t *testing.T) {
+	tests := []struct {
+		name      string
+		images    []string
+		wantImage string
+		wantOK    bool
+	}{
+		{name: "no containers"},
+		{name: "multiple containers", images: []string{"a", "b"}},
```
akavel-reply: 1. please expand to multiline
2. please add explicit `wantOK: false` to make easier to understand

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+	nodes, relations := codeownersClaims(repoNodeID, handles, make(map[pluginapi.NodeID]bool))
+
+	assert.ElementsMatch(t, []pluginapi.NodeClaim{
+		{ID: pluginapi.NodeID{Kind: pluginapi.NodeKindOwner, Path: "github.com/@acme/team"}},
+		{ID: pluginapi.NodeID{Kind: pluginapi.NodeKindOwner, Path: "github.com/@alice"}},
+	}, nodes)
+	assert.ElementsMatch(t, []pluginapi.RelationClaim{
+		{Kind: pluginapi.RelationKindOwnedBy, From: repoNodeID, To: pluginapi.NodeID{Kind: pluginapi.NodeKindOwner, Path: "github.com/@acme/team"}},
+		{Kind: pluginapi.RelationKindOwnedBy, From: repoNodeID, To: pluginapi.NodeID{Kind: pluginapi.NodeKindOwner, Path: "github.com/@alice"}},
+	}, relations)
```
akavel-reply: 1. please use helper func `nodeID` with explicit string Kind values
2. please expand RelationClaim structs to multiline

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func TestCodeownersClaims_Empty(t *testing.T) {
+	nodes, relations := codeownersClaims(pluginapi.NodeID{Kind: pluginapi.NodeKindGitRepository, Path: "github.com/acme/service"}, nil, make(map[pluginapi.NodeID]bool))
```
akavel-reply: line way too long and hard to read

plugins/cmd/github/github_test.go
libera13: Good point, updated to: 
```
{
	name: "skips unsupported encoding and keeps looking",
	responses: map[string]ghContent{
		".github/CODEOWNERS": {Content: codeownersBody, Encoding: "text"},
		"CODEOWNERS":         {Content: base64Body, Encoding: "base64"},
	},
	wantContent:  codeownersBody,
	wantRequests: []string{".github/CODEOWNERS", "CODEOWNERS"},
},
```

plugins/cmd/github/github_test.go
libera13: Tests will behave the same, removed

plugins/cmd/github/main_test.go
libera13: 1. True, I will correct
2. There was test "multiple containers: nothing produced" I will change name to indicate more clearly that it's about multiple images

plugins/cmd/fluxcd/main_test.go
libera13: I couldn't use `nodeID` func as it adds clusterid in prefix. I added new func `rawNodeId` alongside `nodeID` which do not add any prefixes

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,338 @@
...
+			wantRelations: []pluginapi.RelationClaim{
+				{Kind: "owned_by",
+					From: repoA,
+					To:   ownerTeam,
+				},
+				{Kind: "built_from",
+					From: deployment("app"),
+					To:   repoA,
+				},
+			},
```
akavel-reply: The way I'd prefer it look for readability:

```suggestion
			imagesByDeployment: map[string][]string{
				"app": {"ghcr.io/naira-project/service-a:v1"},
			},
			wantNodes: []pluginapi.NodeClaim{
				{ID: repoA,
					Properties: pluginapi.PropertyMap{
						"url": "https://github.com/naira-project/service-a",
						"language": "Go"}},
				{ID: ownerTeam},
				{ID: deployment("app")},
			},
			wantRelations: []pluginapi.RelationClaim{
				{Kind: "owned_by",
					From: repoA,
					To:   ownerTeam,
				},
				{Kind: "built_from",
					From: deployment("app"),
					To:   repoA,
				},
			},
```

plugins/cmd/github/main_test.go
```diff
@@ -0,0 +1,338 @@
...
+			gotImage, gotOK := singleContainerImage(Deployment{Images: tt.images})
+			assert.Equal(t, tt.wantImage, gotImage)
+			assert.Equal(t, tt.wantOK, gotOK)
+		})
+	}
+}
+
+func TestCodeownersClaims(t *testing.T) {
+	repoNodeID := nodeID("git_repository", "github.com/acme/service")
+	handles := []string{"@acme/team", "@alice"}
```
akavel-reply: github [apparently also allows emails](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), so we could test them here and in the parser (`docs@example.com` is the example they give)

akavel-reply (approved):
A few more suggestions, but not critical.

PR #168, opened 2026-08-27
ci: check UI in PR

libera13:
## Description

Test if UI lint works in CI


libera13:
looks I need to test with main

PR #171, opened 2026-08-28
build: deploy keycloak in declarative way

libera13:
## Description

Previously, deploying Keycloak through Flux wasn't possible. It was only possible by manually running tasks from taskfiles, which overwrote Keycloak related envs in catalog and portal.

With this PR, Keycloak deployment is possible with declarative files, which makes it possible to deploy the catalog and portal with Keycloak through Flux or Tilt.

## Type of Change

## How Has This Been Tested?
- [x] task core:deploy - works
- [x] tilt up - works

### DIFF
--- a/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
+++ b/deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
@@ -279,12 +279,11 @@ spec:
             - name: PORT
               value: "8090"
             - name: KEYCLOAK_BASE_URL
-              value: __KEYCLOAK_BASE_URL__
-            # KEYCLOAK_REALM variable must be changed in tasks/keycloak.yml
+              value: http://keycloak.keycloak.svc.cluster.local:8080
             - name: KEYCLOAK_REALM
-              value: __KEYCLOAK_REALM__
+              value: naira
             - name: KEYCLOAK_ISSUER
-              value: __KEYCLOAK_ISSUER__
+              value: http://localhost:8080/realms/naira
             - name: PLUGIN_ADDRESSES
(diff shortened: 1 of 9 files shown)


deploy/dev/stacks/core/infra/kubernetes/catalog.yaml
```diff
@@ -279,12 +279,11 @@ spec:
...
             - name: KEYCLOAK_BASE_URL
-              value: __KEYCLOAK_BASE_URL__
-            # KEYCLOAK_REALM variable must be changed in tasks/keycloak.yml
+              value: http://keycloak.keycloak.svc.cluster.local:8080
             - name: KEYCLOAK_REALM
-              value: __KEYCLOAK_REALM__
+              value: naira
             - name: KEYCLOAK_ISSUER
-              value: __KEYCLOAK_ISSUER__
+              value: http://localhost:8080/realms/naira
```
kocakkaan, line 286: To make it more configurable, I had provided these values previously. I don't know any other easier alternative but do you think putting them hardcoded would be alright for now?

libera13: I'm not sure what you mean.
For me, hardcoded means declared in code, so to change the value you need to change the code. Here, values are passed as env variables. And let's say the same effort is required to change them as previously. Before, when you wanted to change them, you had to modify the taskfile. Now, you need to modify the k8s manifest. From this perspective, it looks the same to me. However, the approach with a declarative file is a big winner for me, as they don't need scripts to be run manually,  tools can deploy files by themselves.

kocakkaan (approved):
Works nice locally as well. Looks good to me 👍

jagdeepm (approved):

PR #173, opened 2026-09-01
feat: add plugin scheduling and display schedules in UI

libera13:
## Description

This PR adds plugin scheduling support across the catalog backend and UI. It exposes configured schedules through the API, runs plugins automatically according to their cron expressions, and displays a human-readable schedule in the Plugins page.

## Type of Change

## How Has This Been Tested?

- [x] `go test ./...` passes

## Screenshots
Screenshot 2026-09-01 at 09 28 11


rootlsz (approved):
looks good 👏

catalog/cmd/catalog/config.go
```diff
@@ -21,10 +22,18 @@ type config struct {
...
+	ShutdownTimeout    time.Duration `env:"SHUTDOWN_TIMEOUT" default:"5s"`
+	// PluginAddresses is a list of plugin name-to-address mappings.
+	// Values must be in the format: name=address (comma-separated).
+	// 	Example: PLUGIN_ADDRESSES="litellm=localhost:9001,mlflow=localhost:9002"
+	PluginAddresses []string `env:"PLUGIN_ADDRESSES"`
+	// PluginSchedules is an optional list of plugin cron schedules.
+	// Values must be in the format: name=cron_expression (comma-separated).
+	// A missing entry means the plugin is manual-only.
+	// 	Example: PLUGIN_SCHEDULES="litellm=0 * * * *,mlflow=@daily"
+	PluginSchedules         []string      `env:"PLUGIN_SCHEDULES"`
```
akavel-reply: I don't like two separate lists keyed by the same names. Starts to be too weird and messy and error-prone (e.g. inevitable accidental typos and other types of desyncing). Maybe it's the time to migrate this to classic JSON/YAML/TOML/... already? See e.g.: https://github.com/naira-project/architecture/blob/main/rfc/rfc-004-core-api-v1-design.md#plugins_registryjson

catalog/internal/httpapi/router.go
```diff
@@ -11,12 +11,13 @@ import (
...
 	"github.com/naira-project/naira/catalog/internal/catalog"
 	"github.com/naira-project/naira/catalog/internal/pluginrun"
+	"github.com/naira-project/naira/catalog/internal/scheduling"
 )
 
 // NewRouter wires up the catalog HTTP API.
 // catalogService serves read-only graph queries (nodes/relations);
 // runner drives and reports on asynchronous plugin runs (plugins/operations).
-func NewRouter(catalogService *catalog.Service, runner *pluginrun.Runner, logger *log.Logger, kc keycloak.Config) (http.Handler, error) {
+func NewRouter(catalogService *catalog.Service, runner *pluginrun.Runner, scheduler *scheduling.Scheduler, logger *log.Logger, kc keycloak.Config) (http.Handler, error) {
```
akavel-reply: (optional) FWIW, at some point I expect it could be worth to start using https://github.com/google/wire (or some fork) - might not be yet, but just sharing the tool link for now, I used it in the past and liked

catalog/internal/httpapi/router.go
```diff
@@ -38,6 +39,8 @@ func NewRouter(catalogService *catalog.Service, runner *pluginrun.Runner, logger
 
 		r.Post("/plugins:run", newRunAllPluginsHandler(runner))
 		r.Get("/plugins", newListPluginsHandler(runner))
+		r.Get("/schedules", newListSchedulesHandler(scheduler, logger))
+		r.Get("/{plugin}/schedule", newGetScheduleHandler(scheduler))
 		r.Post("/{plugin}:run", newRunPluginHandler(runner))
```
akavel-reply: I'd like to propose a different shape of this API:

```
POST /plugins:run - as is

POST /plugins/{plugin}:run
// **Replaces:** POST /{plugin}:run
// I missed this one before somehow; I believe `/{plugin}:run` is not correct
// from AIPs perspective, must have prefix; also anyway `/plugins/{plugin}:run`
// is exactly how this should be joined with existing `/plugins` endpoints AFAIU.

GET /plugins/{plugin}
{
  "name": ...,  // required field per AIPs
  "schedule": "@daily",
  "disabled": true,  // I think "enabled" is the presumed default state, and
                     // can be skipped typically, so maybe we should rather
                     // mark "disabled" as stand-out exception from the
                     // common, default state?
  ...anything else?
}

GET /plugins
[
  { name = "...", schedule = "...", ... },
  { name = "...", ... }
]
// see: https://google.aip.dev/132
// **Replaces:** previous GET /plugins, and also GET /schedules
// (looks like I also missed/overlooked GET /plugins semantics before,
// current ones don't seem valid per AIPs I think)

GET /operation/{operationId}
// Also: it seems to me the resource returned from this endpoint
// doesn't comply with AIP-151

File: catalog/internal/scheduling/memory_store.go [hunk 4]
```diff
@@ -0,0 +1,67 @@
...
+	return result, nil
+}
+
+func (s *MemoryStore) Upsert(schedule Schedule) error {
+	schedule.Plugin = normalizePlugin(schedule.Plugin)
+	if schedule.Plugin == "" {
+		return fmt.Errorf("upsert schedule: %w", ErrInvalidPlugin)
+	}
+	if schedule.UpdatedAt.IsZero() {
+		schedule.UpdatedAt = time.Now().UTC()
```
akavel-reply: If we agreed to not allow changing plugin schedules through API, then I think we don't need Upsert nor UpdatedAt anymore, no? And don't need to keep them in MemoryStore - we could just serve them straight from the static data loaded from config, no?

akavel-reply (changes requested):

catalog/internal/scheduling/types.go
```diff
@@ -0,0 +1,18 @@
...
+// Schedule describes the effective periodic trigger for a plugin.
+// An empty Expression means that the plugin is manual-only.
+type Schedule struct {
+	Plugin     string    `json:"plugin"`
+	Expression string    `json:"expression,omitempty"`
+	Enabled    bool      `json:"enabled"`
+	UpdatedAt  time.Time `json:"updatedAt"`
+}
+
+type Store interface {
```
akavel-reply: Do we need interface? Could we just use concrete type for now instead?

catalog/internal/httpapi/schedules.go
```diff
@@ -0,0 +1,65 @@
...
+	NextPageToken string                `json:"nextPageToken,omitempty"`
+	TotalSize     int32                 `json:"totalSize"`
+}
+
+var scheduleListOptionsSpec = listOptionsSpec{
+	scope:         "schedules",
+	allowedFields: map[string]bool{},
+}
+
+func sortSchedules(schedules []scheduling.Schedule) {
```
akavel-reply: do we need wrapper func? it's short and seems only used in one place, seems to me it would be ok to inline it

catalog/cmd/catalog/config.go
libera13: I've added support for YAML. Thanks for suggestion

catalog/internal/httpapi/router.go
libera13: Added
/plugins/{plugin}
changed
/{plugin}:run -> plugins/{plugin}:run
removed "enabled"/"disabled" field. Empty schedule means disabled schedule.

Unified schedules with plugins, current structure
```
/plugins
[
  { name = "...", schedule = "@daily", ... },
  { name = "...", schedule = "", ... }
]
```

_GET /operation/{operationId}
// Also: it seems to me the resource returned from this endpoint
// doesn't comply with AIP-151_ 
-> for this I will make separate PR https://github.com/naira-project/naira/issues/178

catalog/internal/scheduling/memory_store.go
libera13: Yes, I've removed that code

catalog/internal/catalog/types.go
```diff
@@ -1,6 +1,47 @@
...
+)
+
+// PluginDefinition is the immutable configuration for a registered plugin.
+// An empty Schedule means that the plugin is manually triggered only.
+type PluginDefinition struct {
+	Address  string
+	Schedule string
+}
+
+type PluginConfig map[string]PluginDefinition
```
akavel-reply: "PluginConfig" suggests singular plugin & singular config; the type definition however seems to be for multiple plugins; I'd suggest using some kind of name here that more clearly shows this is plural; ideally, also showing that this is a mapping. Also, I'm confused why the entry has name ...Definition, but the map type has name ...Config - so is it config, or definition? those should be consistent.

catalog/internal/catalog/types.go
```diff
@@ -1,6 +1,47 @@
...
+var ErrInvalidPluginConfig = errors.New("invalid plugin configuration")
+
+func (c PluginConfig) Validate() error {
+	seen := make(map[string]string, len(c))
+	for name, definition := range c {
+		normalizedName := strings.ToLower(strings.TrimSpace(name))
+		if normalizedName == "" {
+			return fmt.Errorf("plugin name %q: %w", name, ErrInvalidPluginConfig)
+		}
+		if previous, exists := seen[normalizedName]; exists {
```
akavel-reply: Maybe we should just reject when `normalizedName != name`? that would be simpler to understand, and thus probably less error-prone in longer term

catalog/internal/catalog/types.go
```diff
@@ -1,6 +1,47 @@
...
+			return fmt.Errorf("plugin names %q and %q normalize to the same name: %w", previous, name, ErrInvalidPluginConfig)
+		}
+		seen[normalizedName] = name
+		if strings.TrimSpace(definition.Address) == "" {
```
akavel-reply: if we're doing "Validate", should we actually also parse the address as an URL here and see if it's valid?

catalog/internal/catalog/types.go
```diff
@@ -1,6 +1,47 @@
...
+			return fmt.Errorf("plugin %q has no address: %w", name, ErrInvalidPluginConfig)
+		}
+		if definition.Schedule != "" {
```
akavel-reply: should we also reject if `defn.Schedule != strings.TrimSpace(defn.Schedule)`?

catalog/internal/httpapi/plugins.go
```diff
@@ -1,22 +1,71 @@
...
+	NextPageToken string           `json:"nextPageToken,omitempty"`
+	TotalSize     int32            `json:"totalSize"`
+}
+
+var pluginListOptionsSpec = listOptionsSpec{
+	scope:         "plugins",
+	allowedFields: map[string]bool{},
+}
+
+func pluginResource(name string, definition catalog.PluginDefinition) PluginResource {
```
akavel-reply: the idiomatic way to write this func name would be:

```suggestion
func newPluginResource(name string, definition catalog.PluginDefinition) PluginResource {
```

catalog/internal/httpapi/plugins.go
```diff
@@ -1,22 +1,71 @@
...
+		resources := make([]PluginResource, 0, len(definitions))
+		for _, name := range slices.Sorted(maps.Keys(definitions)) {
+			resources = append(resources, pluginResource(name, definitions[name]))
+		}
+
+		page, nextPageToken, totalSize, err := paginate(resources, options.pageSize, options.offset, "plugins", logger)
+		if err != nil {
+			return fmt.Errorf("paginating plugins: %w", err)
+		}
+		writeJSON(w, http.StatusOK, ListPluginsResponse{Plugins: page, NextPageToken: nextPageToken, TotalSize: int32FromCount(totalSize, logger)})
```
akavel-reply: Please split the struct to multiline for slightly better readability:

```suggestion
		writeJSON(w, http.StatusOK, ListPluginsResponse{
			Plugins: page,
			NextPageToken: nextPageToken,
			TotalSize: int32FromCount(totalSize, logger),
		})
```

catalog/internal/httpapi/plugins.go
```diff
@@ -1,22 +1,71 @@
...
+var pluginListOptionsSpec = listOptionsSpec{
+	scope:         "plugins",
+	allowedFields: map[string]bool{},
+}
+
+func pluginResource(name string, definition catalog.PluginDefinition) PluginResource {
+	return PluginResource{Name: name, Schedule: definition.Schedule}
+}
+
+// GET /v1/plugins lists configured plugin resources and their schedules.
```
akavel-reply: Please don't write `GET /v1/plugins` here. This function does not control on which endpoint it is attached, so as soon as someone changes the actual code in router, this comment will become stale and misleadling. And we cannot expect somebody changing the router's code to know that we want them to track some arbitrary references in some godocs somewhere in the codebase.

Please also fix this in other funcs in this file.

catalog/internal/httpapi/plugins.go
```diff
@@ -1,22 +1,71 @@
...
 		return nil
 	})
 }
 
+// GET /v1/plugins/{plugin} returns one configured plugin resource.
+func newGetPluginHandler(definitions catalog.PluginConfig) http.HandlerFunc {
+	return handle(func(w http.ResponseWriter, r *http.Request) error {
+		plugin := chi.URLParam(r, "plugin")
+		definition, ok := definitions[plugin]
+		if ok {
```
akavel-reply: The idiomatic and expected way to write this func would be to use `if !ok { ... }` block here.

catalog/internal/scheduling/scheduler.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/robfig/cron/v3"
+)
+
+var (
+	ErrInvalidPlugin = errors.New("invalid plugin name")
+)
+
+type RunStarter interface {
+	RunPluginAsync(ctx context.Context, plugin string) (operations.Operation, error)
+}
```
akavel-reply: 1. Do we need a whole interface, or could we just use a function pointer type?
2. Could it make more sense to name the type "PluginRunStarter"? I don't understand why we drop the "Plugin" part here - isn't this about plugins? Or is it expected to be some generic starter of whatever non-plugin things too? but it takes a name of a plugin through args, so it seems it's only for plugins in the end?

catalog/internal/scheduling/scheduler.go
```diff
@@ -0,0 +1,74 @@
...
+	return sch, nil
+}
+
+func (s *Scheduler) registerSchedules(config catalog.PluginConfig, starter RunStarter, logger *log.Logger) error {
+	for plugin, definition := range config {
+		expr := definition.Schedule
+		if plugin == "" {
+			return ErrInvalidPlugin
+		}
+		if expr == "" {
```
akavel-reply: I think it could be better to create a const for this special value, and use it here as well as in the validator.

catalog/cmd/catalog/config_test.go
```diff
@@ -0,0 +1,75 @@
...
+`,
+			createFile: true,
+			wantPlugins: catalog.PluginConfig{
+				"mlflow": {
+					Address:  "localhost:50051",
+					Schedule: "*/5 * * * *",
+				},
+			},
+		},
+	}
```
akavel-reply, line 53: Given that the config loader also does validation of the contents, I think it could make sense to do some tests of Validate here.

catalog/cmd/catalog/config_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/naira-project/naira/catalog/internal/catalog"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoadPluginConfig(t *testing.T) {
+	tests := []struct {
+		name        string
+		config      string
+		createFile  bool
```
akavel-reply: (optional) You could revert this variable (e.g. `dontWriteFile`, or `skipFileCreation`), so that the default `false` would mean writing the file. Then, only one test case would need to set it explicitly to `true`, rather than all except one having to set it explicitly to `true` as is now.

catalog/internal/scheduling/scheduler_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type stubStarter struct {
+	calls chan string
+	err   error
+}
+
+func (s *stubStarter) RunPluginAsync(_ context.Context, plugin string) (operations.Operation, error) {
+	if s.calls != nil {
```
akavel-reply: is this check needed? you seem to always create this type with non-empty `calls` field - this looks like dead code to me

catalog/internal/scheduling/scheduler_test.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+			scheduler, err := NewConfiguredScheduler(config, starter, log.New(io.Discard, "", 0))
+
+			if tt.wantErr {
+				require.Error(t, err)
+				if tt.expectedErr != nil {
+					assert.ErrorIs(t, err, tt.expectedErr)
+				}
+				if tt.errContains != "" {
+					assert.Contains(t, err.Error(), tt.errContains)
```
akavel-reply: I think we could shorten to `assert.ErrorContains(...)`

catalog/internal/catalog/types.go
libera13: Thanks for suggestion. I applied `net.SplitHostPort(config.Address)` from what I read it fits more here

catalog/internal/scheduling/scheduler.go
libera13: Ok, I used function pointer type and `RunPluginFunc` name

catalog/internal/catalog/types.go
```diff
@@ -3,40 +3,56 @@ package catalog
...
-			if _, err := cron.ParseStandard(definition.Schedule); err != nil {
-				return fmt.Errorf("plugin %q has invalid schedule %q: %w", name, definition.Schedule, ErrInvalidPluginConfig)
+
+		if strings.ToLower(strings.TrimSpace(name)) != name {
+			return fmt.Errorf("plugin name %q must be lowercased without leading/trailing whitespace: %w", name, ErrInvalidPluginConfig)
+		}
+		if config.Schedule != strings.TrimSpace(config.Schedule) {
+			return fmt.Errorf("plugin %q schedule %q must not have leading or trailing whitespace: %w", name, config.Schedule, ErrInvalidPluginConfig)
+		}
+		if config.Address != strings.TrimSpace(config.Address) {
```
akavel-reply: 1. this checks Address - would be nice to move it closer to another check of Address that is further above
2. (optional) there's a trick I sometimes like to use in similar code, which makes it a bit shorter - which is to group it into a `switch` block:

```
switch {
case name == "":
    return fmt.Errorf(...)
case name != strings.ToLower(...):
    return fmt.Errorf(...)
case config.Address == "":
    return fmt.Errorf(...)
case config.Address != strings.TrimSpace(config.Address):
    return fmt.Errorf(...)
...
}

/* the more complex checks usually need to still stay outside though: */
host, port, err := ...
if ... 
```

catalog/internal/catalog/types.go
```diff
@@ -3,40 +3,56 @@ package catalog
...
-		seen[normalizedName] = name
-		if strings.TrimSpace(definition.Address) == "" {
+		if config.Address == "" {
 			return fmt.Errorf("plugin %q has no address: %w", name, ErrInvalidPluginConfig)
 		}
-		if definition.Schedule != "" {
-			if _, err := cron.ParseStandard(definition.Schedule); err != nil {
-				return fmt.Errorf("plugin %q has invalid schedule %q: %w", name, definition.Schedule, ErrInvalidPluginConfig)
+
+		if strings.ToLower(strings.TrimSpace(name)) != name {
```
akavel-reply: this if's condition has untypical order - I'd prefer to keep it consistent with usual way of doing things, i.e. `if name != ...`. Also, please move it closer to the other check of `name` which is higher above.

catalog/cmd/catalog/config.go
```diff
@@ -68,27 +68,27 @@ func loadConfig() (config, error) {
 	}, nil
 }
 
-func loadPluginConfig(path string) (catalog.PluginConfig, error) {
+func loadPluginConfig(path string) (catalog.PluginConfigsByName, error) {
```
akavel-reply: Should we call this function `loadAndValidatePluginConfig`? (maybe let's have a call)

catalog/cmd/catalog/config_test.go
```diff
@@ -0,0 +1,187 @@
...
+				assert.ErrorContains(t, err, tt.wantErrText)
+			} else {
+				require.NoError(t, err)
+				assert.Equal(t, tt.wantPlugins, plugins)
+			}
+		})
+	}
+}
+
+func TestPluginConfigsByName_Validate(t *testing.T) {
```
akavel-reply, line 75: My intention in the comment to previous review was for this to be part of the `loadPluginConfig` test. Now, we're testing what Validate checks - which generally is good and theoretically very "unit-testing'y" - but I'm worried we're not testing whether `Validate()` is actually called by `loadPluginConfig`... so, Validate might be all fine and cool, but we could still allow invalid configs by never calling it at all.

catalog/cmd/catalog/config_test.go
```diff
@@ -0,0 +1,187 @@
...
+		},
+		{
+			name: "fails when schedule cron format is invalid",
+			plugins: catalog.PluginConfigsByName{
+				"mlflow": {
+					Address:  "localhost:50051",
+					Schedule: "@Daily",
+				},
+			},
+			wantErr: true,
```
akavel-reply: Could we check the error message, to make sure the error is actually for the reason we think it is?

catalog/internal/scheduling/scheduler.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	}
+	return nil
+}
+
+func (s *Scheduler) Stop(ctx context.Context) error {
+	cronCtx := s.cron.Stop()
+	select {
+	case <-cronCtx.Done():
+		return nil
```
akavel-reply, line 68: Is it important for us to wait for the jobs to finish? if yes, why?

catalog/cmd/catalog/config_test.go
libera13: Understood. In `loadPluginConfig`, there's a case that expects an error when there is no address, which is one of the validation checks. I can rename this test case to be even more directed toward what you described, changing it from "returns an error when a plugin has no address" to "check if validation is triggered"

catalog/internal/scheduling/scheduler.go
libera13: For me it was easier to reason about the code. Stop http server, stop scheduler. Wait for ongoing to finish. 
SIGINT,SIGTERM anyway ends context for plugin calls, which stops plugins. So I thought let's all state to be updated correctly. 

after reading your message I spotted the logic error that in current code it doesn't wait anyway for jobs to finish as context is cancelled beforehand and second case is always used. Will fix that

akavel-reply (approved):

PR #175, opened 2026-09-02
feat: tech radar implementation

### LINKED ISSUE #161: Implementing Tech Radar
## Overview

**User Story:** As **Platform Peter**, I want to publish our organization's technology decisions which models, agentic patterns, and knowledge techniques to adopt, trial, assess, or hold as a versioned configuration file, so that governance decisions are visible inside Naira instead of living in a separate wiki or slide deck.

**User Story:** As **AI Engineer rAIner**, I want to see which technologies our AI board has endorsed and jump from a radar entry directly to its catalog entry, so that I pick approved building blocks without asking around.

**User Story:** As **App Dev Abdel**, I want defaults and deprecations shown with their rationale, so that I know what to use in my application right now and what to migrate away from.

The Tech Radar is a read-only governance view rendered from an admin-managed configuration file. It puts organizational decisions next to the assets the catalogs already describe.

---

## Context and Problem Statement

Naira's catalogs answer *what exists*: models, MCP servers, datasets, inference endpoints, software components. None of them answers *what the organization endorses*. That decision layer today lives in wikis, review-board minutes, and slides and are so disconnected from the assets it governs, and invisible at the moment an engineer picks a model or a retrieval technique.

The technology radar is an established format for exactly this: entries placed in quadrants (topic areas) and rings (adoption stages), with movement markers and a recorded rationale per entry. Bringing it into Naira means a radar entry for a model can link straight to that model's catalog entry, so the decision and the asset are one click apart.

A deliberate constraint shapes this feature: the radar is **not** edited in the portal. Admins manage a declarative configuration file, reviewed and changed through whatever process the organization already uses for configuration. The portal renders it read-only. This keeps governance auditable and keeps Naira out of the business of being an editing tool for board decisions.

---

### 1. Radar configuration file

A single YAML file defines the whole radar: radar metadata (title, edition label, owner, review cadence, ring definitions text), an ordered list of quadrants (id, name), an ordered list of rings (id, name, description), and the entries. Each entry carries `id`, `name`, `quadrant`, `ring`, `moved` (`in` / `out` / `none`), `owner`, `rationale`, and an optional `catalog_ref` (`kind` + `path`) pointing at an existing catalog node. Quadrant and ring taxonomies are fully configurable; the defaults from the design drafts (Models / Agentic / Knowledge / Others, adopt / trial / assess / hold) ship as the documented example. The schema carries a `schema_version` field.

### 2. Loading and validation

Admins provide the file as a mounted file or Kubernetes ConfigMap read by the radar plugin. The file is re-read each sync cycle, so changes appear without redeployment. It is validated against the schema before use; an invalid file is rejected with precise errors (line, field, reason) written to the plugin log, and the last known good radar remains in place.

### 3. Radar plugin

A standard plugin implementing the existing `pluginapi.Plugin` contract. `Collect()` emits one `tech_radar` node carrying the taxonomy and edition metadata, one `tech_radar_entry` node per entry (properties: quadrant, ring, moved, owner, rationale, edition), and a `references` relation from each entry with a `catalog_ref` to the target catalog node. Snapshot reconciliation prunes entries removed from the file. No catalog core changes.

### 4. API

Radar data is served by the existing catalog REST API: `GET /v1/nodes?filter=kind="tech_radar_entry"` and the `references` relations. No new endpoints.

### 5. UI

An OpenMFP micro-frontend with two views. The **radar overview** renders the classic circle: rings from outer to inner, quadrant sectors, numbered dots colored by ring, movement markers, plus per-quadrant summary cards and the ring definitions from the config. 
The **quadrant detail** view is a table (entry, ring, movement, owner, catalog link, rationale) with filters for ring, moved-only, and has-catalog-entry. Entries with a resolved `references` relation link to the target's catalog detail page.

---

## Functional Requirements

1. The configuration file is the sole source of truth; the portal offers no write path.
2. Quadrants and rings (names, order, descriptions) are defined in the config file.
3. Every entry carries name, quadrant, ring, movement, owner, and rationale.
6. The radar overview visualizes all entries by quadrant and ring with movement markers.
7. Each quadrant has a tabular detail view filterable by ring, moved-only, and has-catalog-entry.
8. The current edition label and a summary of moved entries are displayed.
9. Config changes take effect on the next sync cycle without redeployment.
10. Radar entries are queryable through the standard catalog API.

---

## Non-Functional Requirements

- **Plugin-only:** no changes to the catalog core; schema additions go through `pluginapi/schema.go` as for other entity kinds.
- **Safety:** a malformed config must never blank the radar; last known good state persists.
- **Documentation:** the config schema is documented with a complete annotated example.

---

## Acceptance Criteria

- [ ] **Given** a valid config file is mounted, **when** the plugin syncs, **then** the radar renders the configured quadrants, rings, and entries.
- [ ] **Given** an admin mounts an invalid config, **when** the plugin syncs, **then** the previous radar remains visible and validation errors name the offending field and line.
- [ ] **Given** entries are marked `moved: in` or `moved: out`, **when** the overview loads, **then** the moved count and per-entry markers are shown.
- [ ] **Given** the moved-only filter is active, **when** the table renders, **then** only moved entries are listed.
- [ ] **Given** an admin renames a ring in the config, **when** the next sync completes, **then** the UI reflects the new taxonomy without redeployment.
- [ ] **Given** a client queries `GET /v1/nodes?filter=kind="tech_radar_entry"`, **then** paginated entries with all radar properties are returned.
- [ ] **Given** the user hovers on the radar over the marker of a technology, a infobox opens with the rational and further information.

---

## Out of Scope

- In-portal editing or proposal workflows (the "Propose entry / change" buttons from the drafts) — a possible later feature.
- Browsing or diffing historic editions; only the current edition and its declared movement markers are shown.
- Enforcement of radar decisions (e.g., the gateway blocking "hold" models).
- Automated ring recommendations from usage or observability data.
- Multiple radars or workspace-scoped radars; one radar per Naira instance for now.
- Sourcing the config from a Git repository — the mounted-file mechanism composes with any external GitOps tooling, but Git sync is not part of this feature.

mkorbi:
## Description

This PR introduces a tech radar as plugin for Naira. It uses a ConfigMap for configuration.
Implements https://github.com/naira-project/naira/issues/161.

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📝 Documentation update

## How Has This Been Tested?

- [X] `go test ./...` passes
- [ ] `helm lint` passes
- [X] Manual testing on Kubernetes

## Checklist

- [X] My code follows the project's coding standards ([AGENTS.md](AGENTS.md))

 Describe how and for what AI was used in this PR
## AI Usage

AI was used for the development and testing of this feature.

Screenshot 2026-09-02 at 14 05 41
Screenshot 2026-09-02 at 14 05 55
Screenshot 2026-09-02 at 14 05 47


mkorbi:
Thinking about, that this doesn't necessarly make sense as a standalone plugin with yet another container. I could keep the tech radar logically a plugin but run it in-process inside the catalog binary instead of as a gRPC sidecar. 

As the pluginapi is currently gRPC-only (right?), this would mean adding a small in-process plugin registry to the plugin manager. 
However, that could be interesting to have too, for future built-in collector that could benefit from. What we would get are:

- No tech-radar:local image, no sidecar container, no PLUGIN_ADDRESSES entry.
- An on/off flag (e.g. TECH_RADAR_ENABLED env var, set from Helm values) that simply registers the in-process plugin or doesn't.
- Zero behavior change: same node kinds, same sync trigger, same error surfacing, same last-known-good

WDYT?

plugins/cmd/tech_radar/nodes.go
```diff
@@ -0,0 +1,94 @@
...
+	propertyKeyRings         = "rings"
+	propertyKeyEntryCount    = "entry_count"
+	propertyKeyMovedCount    = "moved_count"
+	propertyKeyQuadrant      = "quadrant"
+	propertyKeyRing          = "ring"
+	propertyKeyMoved         = "moved"
+	propertyKeyRationale     = "rationale"
+	propertyKeyIndex         = "index"
+	propertyKeyRadar         = "radar"
+)
```
akavel-reply, line 26: with so many items, maybe we could keep them sorted alphabetically?

plugins/cmd/tech_radar/main.go
```diff
@@ -0,0 +1,68 @@
+// tech_radar plugin that renders an admin-managed technology radar
```
akavel-reply: small grammar typo

```suggestion
// tech_radar plugin renders an admin-managed technology radar
```

plugins/cmd/tech_radar/main.go
```diff
@@ -0,0 +1,68 @@
...
+// configuration file into the catalog. The YAML file is the sole source of
+// truth: it defines the radar metadata, quadrant and ring taxonomies, and the
+// entries with their adoption ring, movement marker, owner, and rationale.
+//
+// The file is re-read on every collect, so ConfigMap updates take effect on
+// the next sync without a redeployment. An invalid file fails the collect
+// with errors naming the offending line and field, which keeps the previous
+// radar snapshot in place (the catalog never applies a failed run).
```
akavel-reply: Small verbiage changes: plugin has no power to enforce anything in the catalog, so it can't make claims what the catalog "does" - if catalog changes behavior, this comment will become false and misleading. Instead, plugin should explain its assumptions, the contract it requires - this can be phrased e.g. with "should":

```suggestion
// with errors naming the offending line and field, which should keep the previous
// radar snapshot in place (the catalog should never apply a failed run).
```

plugins/cmd/tech_radar/config_test.go
```diff
@@ -0,0 +1,281 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const validConfig = `schema_version: 1
```
akavel-reply: Maybe we could put this into a `sample_config.yaml` file in the plugin's directory, to help users find it and see how a file should look like? (The test would then read this file instead of reading from the const)

plugins/cmd/tech_radar/nodes.go
```diff
@@ -0,0 +1,94 @@
...
+
+const maxTextLength = 2000
```
akavel-reply: Please add a godoc comment clarifying what "text" this refers to

plugins/cmd/tech_radar/nodes.go
```diff
@@ -0,0 +1,94 @@
...
+
+// buildResponse turns a validated radar config into catalog claims: one
+// tech_radar node carrying the taxonomy plus one tech_radar_entry node per
+// entry. The radar emits no relations by design — it describes the target
+// vision, not what is currently deployed.
+func (p *Plugin) buildResponse(cfg *radarConfig) pluginapi.CollectResponse {
+	radarID := pluginapi.NodeID{Kind: pluginapi.NodeKindTechRadar, Path: cfg.Radar.ID}
+
+	movedCount := 0
```
akavel-reply, line 46: What's this all "moved" thing about? I'm confused and don't understand

plugins/cmd/tech_radar/nodes.go
```diff
@@ -0,0 +1,94 @@
...
+			propertyKeySchemaVersion: strconv.Itoa(cfg.SchemaVersion),
+			propertyKeyQuadrants:     string(quadrantsJSON),
+			propertyKeyRings:         string(ringsJSON),
+			propertyKeyEntryCount:    strconv.Itoa(len(cfg.Entries)),
+			propertyKeyMovedCount:    strconv.Itoa(movedCount),
+		},
+	})
+
+	for i, e := range cfg.Entries {
+		rationale := truncate(e.Rationale)
```
akavel-reply: why truncate this one specific field? why truncate at all, and if yes, then why not truncate others too?

plugins/cmd/tech_radar/config.go
```diff
@@ -0,0 +1,276 @@
...
+)
+
+const supportedSchemaVersion = 1
+
+const defaultRadarID = "default"
+
+const (
+	movedIn   = "in"
+	movedOut  = "out"
+	movedNone = "none"
```
akavel-reply, line 20: How do you plan to communicate to the writer of the radar YAML file that those are the values they can use in it? also, even for me it's not immediately obvious where those values can be used, and what they mean.

Are we using some standard format of "tech radar" data? if yes, is there a link to a spec we could provide e.g. in the Readme?

plugins/cmd/tech_radar/config.go
```diff
@@ -0,0 +1,276 @@
...
+	Quadrant  string `yaml:"quadrant"`
+	Ring      string `yaml:"ring"`
+	Moved     string `yaml:"moved"`
+	Owner     string `yaml:"owner"`
+	Rationale string `yaml:"rationale"`
+
+	line int
+}
+
+func (c *radarConfig) UnmarshalYAML(node *yaml.Node) error {
```
akavel-reply, line 82: I'm not super convinced it's worth adding all those functions just to be able to give the user a line number. But even if we want this, I think it should be possible to create a helper generic function, that would be called from each of those functions, so that they'd become practically oneliners.

plugins/cmd/tech_radar/config.go
```diff
@@ -0,0 +1,276 @@
...
+	if err := checkKnownFields(node, "radar config", "schema_version", "radar", "quadrants", "rings", "entries"); err != nil {
+		return err
```
akavel-reply: Please don't return "naked errors". See: https://github.com/naira-project/naira/blob/main/AGENTS.md#go-error-handling

plugins/cmd/tech_radar/config.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	type alias radarConfig
+	var decoded alias
+	if err := node.Decode(&decoded); err != nil {
+		return err
```
akavel-reply: As above - please don't return "naked errors". (Same also in all the other functions.)

docs/tech-radar.md
akavel-reply, line 1: I found this file accidentally. Please make sure the plugin's Readme links to this.

Truth said, personally I'd move this whole file to the plugin's directory.

docs/tech-radar.md
```diff
@@ -0,0 +1,135 @@
...
+  previous radar stays visible ("last known good"). The errors appear in the
+  plugin log and in the sync operation's error message on the Plugins page.
+- Radar data is served by the standard catalog API — no dedicated endpoints:
+
+  ```
+  GET /v1/nodes?filter=kind="tech_radar"
+  GET /v1/nodes?filter=kind="tech_radar_entry"
+  ```
+
+Note that the catalog store is in-memory: after a catalog restart the radar
```
akavel-reply: This note should be removed. It's an artifact of the current implementation of the catalog, it's not planned to stay so for all the time. Even if, it's not the right place to explain it, it's a characteristic of the catalog as a whole.

docs/tech-radar.md
```diff
@@ -0,0 +1,135 @@
...
+(like all catalog data) is empty until the next plugin sync runs.
+
+## Operating the radar
+
+The dev environment mounts the config from the `tech-radar-config` ConfigMap
```
akavel-reply: I find this section confusing. "Operating the radar" sounds production-related, but the section talks about "the dev environment". Who's the intended target reader for this document?

The contents currently sound like counter-productive, troublesome AI output to me.

docs/tech-radar.md
```diff
@@ -0,0 +1,135 @@
...
+| `radar.owner` | yes | Owning team or board. |
+| `quadrants` | yes | Exactly **4** quadrants, in display order. Each has `id` and `name`. |
+| `rings` | yes | 1–6 rings, ordered **innermost to outermost**. Each has `id`, `name`, and an optional `description`. |
+| `entries` | yes | The radar entries (may be empty). |
+| `entries[].id` | yes | Stable identifier; with `radar.id` it forms the node path. |
+| `entries[].name` | yes | Display name. |
+| `entries[].quadrant` | yes | Must match a declared quadrant `id`. |
+| `entries[].ring` | yes | Must match a declared ring `id`. |
+| `entries[].moved` | no | `in`, `out`, or `none` (default). Movement since the previous edition. |
+| `entries[].owner` | yes | Team accountable for the decision. |
```
akavel-reply: for what decision?

plugins/cmd/tech_radar/nodes.go
```diff
@@ -0,0 +1,94 @@
...
+				propertyKeyMoved:     e.Moved,
+				propertyKeyOwner:     e.Owner,
+				propertyKeyRationale: rationale,
+				propertyKeyIndex:     strconv.Itoa(i),
+				propertyKeyRadar:     cfg.Radar.ID,
+			},
+		})
+	}
+
+	return pluginapi.CollectResponse{Nodes: nodes, Relations: []pluginapi.RelationClaim{}}
```
akavel-reply: Hopefully the simpler version should work:


```suggestion
	return pluginapi.CollectResponse{Nodes: nodes, Relations: nil}
```

plugins/cmd/tech_radar/main.go
```diff
@@ -0,0 +1,68 @@
...
+//
+// # Environment Variables
+//
+//   - TECH_RADAR_CONFIG_PATH (optional) - path to the radar YAML file;
+//     defaults to /etc/naira/techradar/radar.yaml.
```
akavel-reply, line 14: Please generate a readme from the comment above with `go generate` - see e.g.:

https://github.com/naira-project/naira/blob/4a0cba8c17612479b2de9f0c055bb2dc882a13a6/plugins/cmd/depl_uses_litellm/plugin.go#L36

plugins/cmd/tech_radar/main_test.go
```diff
@@ -0,0 +1,147 @@
...
+	require.Contains(t, entries, "naira/claude-sonnet")
+	first := entries["naira/claude-sonnet"]
+	assert.Equal(t, "Claude Sonnet", first["title"])
+	assert.Equal(t, "models", first["quadrant"])
+	assert.Equal(t, "adopt", first["ring"])
+	assert.Equal(t, "in", first["moved"])
+	assert.Equal(t, "ml-platform", first["owner"])
+	assert.Equal(t, "Default general-purpose model.", first["rationale"])
+	assert.Equal(t, "0", first["index"])
+	assert.Equal(t, "naira", first["radar"])
```
akavel-reply: Could we just compare the whole `first` with a map of values, rather than comparing field by field manually?

Something like:
```
wantSonnet := map[string]string{
  "title": "Claude Sonnet",
  "quadrant": "models",
  ...
}
assert.Equal(t, wantSonnet, entries["naira/claude-sonnet"])

**Comment by akavel-reply** [comment]
> Thinking about, that this doesn't necessarly make sense as a standalone plugin with yet another container. I could keep the tech radar logically a plugin but run it in-process inside the catalog binary instead of as a gRPC sidecar.

It's an optimization, but it would increase complexity of the catalog, because it'd require adding, handling, and managing, a whole new kind of plugins ("in-process" ones), and maintaining code for them from now on.

I'm fine with considering such optimization at some point in the future, but I'm strongly opposed to it at this moment. For now, it'd require extra implementation effort, and increase complexity of the overall system's code substantially, while I'm not convinced this cost is offset by benefits we'd get from this optimization. Notably, we don't even know if that will be a bottleneck for us in the final system at all. If the system will have e.g. 100+ plugins, then a difference of resources use for one of them seems negligible to me. However, if we'll find that many of them will benefit from this "in-process" model, then it could show up to be worth to do this then, and we'd better understand the performance gains we're fighting for in the big picture.

**Comment by akavel-reply** [comment]
Please ping me on Teams for another review when comments above in Go code are addressed. Please make sure each comment is replied to, either as "implemented", or with a discussion why a different approach is suggested, or why the comment was rejected or postponed to a future linked ticket. Otherwise, I'd have no way of discerning if the comment was consciously rejected or accidentally overlooked.

File: plugins/cmd/tech_radar/config.go (hunk 8 above)
Comment (rootlsz) [reply plugins/cmd/tech_radar/config.go:20]: isn't this part of configuration schema? 👀 
- https://github.com/naira-project/naira/pull/175/changes#diff-bcb53719135fcfce7222284cf7632d80468201c47f1c3ae1902a0df26fd1f41eR72

about the standard format - i think there is only one :D 
- https://www.thoughtworks.com/radar/byor
and you can only move one step at the time. 

Maybe in that case, we should add a "?" button that, when clicked, explains in two sentences the mechanics of the tech radar? In the ui it’s already quite clear what's happening, and the overall idea is described in 
- https://github.com/naira-project/naira/pull/175/changes#diff-bcb53719135fcfce7222284cf7632d80468201c47f1c3ae1902a0df26fd1f41eR3

File: docs/tech-radar.md
Comment (mkorbi) [reply docs/tech-radar.md:1]: moved

File: docs/tech-radar.md (hunk 12 above)
Comment (mkorbi) [reply docs/tech-radar.md]: ok

File: docs/tech-radar.md (hunk 13 above)
Comment (mkorbi) [reply docs/tech-radar.md]: changed

File: docs/tech-radar.md (hunk 14 above)
Comment (mkorbi) [reply docs/tech-radar.md]: Every entry, and every update of the radar falls under one responsibility. The field is there to tell who this is.

File: plugins/cmd/tech_radar/config.go (hunk 11 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/config.go]: fixed

File: plugins/cmd/tech_radar/config.go (hunk 10 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/config.go]: fixed

File: plugins/cmd/tech_radar/config_test.go (hunk 4 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/config_test.go]: good point

File: plugins/cmd/tech_radar/main.go (hunk 16 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/main.go:14]: ok

File: plugins/cmd/tech_radar/main_test.go (hunk 17 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/main_test.go]: adressed

File: plugins/cmd/tech_radar/nodes.go (hunk 6 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/nodes.go:46]: entries in a tech radar are moving over time. So, e.g. there is something fancy new cool, you start with trial or asses, if it proves its value it moves up or in, to adopt (so a strict recommendation to use it) or moves out to hold, or gets entirely off the radar.

File: plugins/cmd/tech_radar/nodes.go (hunk 7 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/nodes.go]: the other fields are by nature not that long, but I will add a truncate to them too to keep consitent. The rational might become way to long, but I prefer to just cut it instead of rejecting it.

File: plugins/cmd/tech_radar/nodes.go (hunk 15 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/nodes.go]: ok

File: plugins/cmd/tech_radar/config.go (hunk 8 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/config.go:20]: It's mentioned in the docs on the website. 

There is no real standard, more like a common sense or approach.

File: plugins/cmd/tech_radar/config.go (hunk 9 above)
Comment (mkorbi) [reply plugins/cmd/tech_radar/config.go:82]: this config is hand-edited YAML behind a ConfigMap and the error's whole journey is: admin edits file → sync fails → message on the Plugins page or so. so there is a huge room for error, and the messages is the only debugging tool the admin gets.

But we can simplify that.

File: ui/src/lib/radarLayout.ts [hunk 18]
```diff
@@ -0,0 +1,221 @@
...
+  for (let i = 0; i < value.length; i++) {
+    hash ^= value.charCodeAt(i);
+    hash = Math.imul(hash, 0x01000193);
+  }
+  return hash >>> 0;
+}
+
+/** Maps a hash to [0, 1). */
+function unitFromHash(hash: number): number {
+  return (hash % 10_000) / 10_000;
```
rootlsz, line 84: probably will never happen, but if hash value is negative (still number type) then result won't be in range [0,1) - maybe changing it is not necessary

rootlsz (approved):
the algorithm for positioning elements inside the radar looks quite complicated at first glance, but in the end it's pretty smart and elegant considering how much it does :D overall it looks good 👍

libera13:
> > Thinking about, that this doesn't necessarly make sense as a standalone plugin with yet another container. I could keep the tech radar logically a plugin but run it in-process inside the catalog binary instead of as a gRPC sidecar.
> 
> It's an optimization, but it would increase complexity of the catalog, because it'd require adding, handling, and managing, a whole new kind of plugins ("in-process" ones), and maintaining code for them from now on.
> 
> I'm fine with considering such optimization at some point in the future, but I'm strongly opposed to it at this moment. For now, it'd require extra implementation effort, and increase complexity of the overall system's code substantially, while I'm not convinced this cost is offset by benefits we'd get from this optimization. Notably, we don't even know if that will be a bottleneck for us in the final system at all. If the system will have e.g. 100+ plugins, then a difference of resources use for one of them seems negligible to me. However, if we'll find that many of them will benefit from this "in-process" model, then it could show up to be worth to do this then, and we'd better understand the performance gains we're fighting for in the big picture.

We could talk about it @akavel-reply. For me, it's even a question if this should be a plugin. Right now let's say its semi-plugin. From a backend perspective, it's only a plugin because it allows us to pass the tech-radar config file to the UI easily. `/nodes` It's used only as a transport layer. And we have a plugin on the backend, which, let's say, makes it logically independent. But from a UI perspective, this code is directly in the `ui/` directory, so there is no logical separation on that side at all. If it will be separated UI code too, then I think we could say it's a full plugin.

The first option for me would be just to make this a standalone UI plugin, without backend. Parsing this config file for the tech-radar could be done in TypeScript.
Pros: The tech-radar really becomes a plugin that can be installed and removed.
Cons: There is no backend, so changing the config would require a rebuild of the UI-plugin. There will be no "last known good", without backend 

Second option. Maybe if it's tightly coupled in the UI, we could also couple it tightly in the backend, and then just create a separate `techradar` package in the `catalog/internal` directory. Instead of trying to pass the tech-radar config through nodes and relations, we could just create new endpoints with a suitable API structure and mount them in `catalog/internal/httpapi/router.go`

plugins/cmd/tech_radar/config.go
```diff
@@ -0,0 +1,327 @@
...
+	e.line = line
+	return nil
+}
+
+// decodeStrict decodes a mapping node into T, rejecting keys that don't match
+// any yaml-tagged field of T so typos surface with their line instead of
+// being silently dropped. It returns the node's line for validation errors.
+func decodeStrict[T any](node *yaml.Node, context string) (T, int, error) {
+	var decoded T
+	if err := checkKnownFields(node, context, yamlFieldNames[T]()); err != nil {
```
akavel-reply: If we're always checking that config doesn't use fields not defined in struct, could we just use https://pkg.go.dev/gopkg.in/yaml.v3#Decoder.KnownFields instead, and then delete all those custom `UnmarshalYAML` methods?

plugins/cmd/tech_radar/main.go
```diff
@@ -0,0 +1,68 @@
...
+//
+// The file is re-read on every collect, so ConfigMap updates take effect on
+// the next sync without a redeployment. An invalid file fails the collect
+// with errors naming the offending line and field, which should keep the previous
+// radar snapshot in place (the catalog should never apply a failed run).
+//
+// # Environment Variables
+//
+//   - TECH_RADAR_CONFIG_PATH (optional) - path to the radar YAML file;
+//     defaults to /etc/naira/techradar/radar.yaml.
```
akavel-reply, line 14: In the [comment in previous review](https://github.com/naira-project/naira/pull/175#discussion_r3924604674), my intention here was to have a `//go:generate ...` line added in this file, so that when modifying the contents of the godoc comment above it would be very easy to automatically regenerate the README and keep it in sync. If we are _not doing_ this (I see the README is now much bigger than this comment), I would suggest reducing this godoc comment to be very minimal, and include a line like: "`For more details, see the README.md file.`". At the current state, we already have the Environment Variables described in two places (here in godoc, but also in the README), which will inevitably lead to them becoming out of sync sooner or later, leading to confusion and mistakes, which will be quite undesirable.

plugins/cmd/tech_radar/sample_config.yaml
```diff
@@ -0,0 +1,48 @@
+# Sample tech radar configuration.
+#
+# Schema reference: plugins/cmd/tech_radar/README.md. The plugin reads the config from
+# /etc/naira/techradar/radar.yaml by default; TECH_RADAR_CONFIG_PATH overrides
```
akavel-reply: (optional) I'd avoid duplicating information across files, as it's notoriously hard to keep such duplication in sync, and it'll inevitably lead to discrepancies and thus can easily increase customer's (and developers') confusion and frustration. The `/etc/naira/techradar/radar.yaml` info and the `TECH_RADAR_CONFIG_PATH` are already explained in README - I'd suggest dropping them from here, and rephrasing the comment around e.g. to:

```
# Schema reference and usage details: plugins/cmd/tech_radar/README.md
```

plugins/cmd/tech_radar/config_test.go
```diff
@@ -0,0 +1,288 @@
...
+			wantErrs: []string{"entries: field is required (use [] for a radar without entries)"},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			_, err := parseRadarConfig([]byte(test.config))
+			require.Error(t, err)
+			for _, want := range test.wantErrs {
+				assert.Contains(t, err.Error(), want)
```
akavel-reply: (minor) can we use `assert.ErrorContains`?

plugins/cmd/tech_radar/main_test.go
```diff
@@ -0,0 +1,170 @@
...
+	result := make(map[string]pluginapi.PropertyMap)
+	for _, node := range response.Nodes {
+		if node.ID.Kind == kind {
+			result[node.ID.Path] = node.Properties
+		}
+	}
+	return result
+}
+
+func TestNewValidatesConfig(t *testing.T) {
```
akavel-reply:
```suggestion
func TestNewValidatesConfigPath(t *testing.T) {
```
"ValidatesConfig" suggested to me we're validating config contents; but this tests seems to validate CONFIG_PATH

plugins/cmd/tech_radar/main_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}, entries["naira/naive-rag"])
+}
+
+func TestCollectFailsWithoutTouchingResponseOnMissingFile(t *testing.T) {
+	p := newTestPlugin(t, filepath.Join(t.TempDir(), "does-not-exist.yaml"), testLogger())
+
+	response, err := p.Collect(t.Context())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "does-not-exist.yaml")
+	assert.Empty(t, response.Nodes, "a failed collect returns a zero-value response")
```
akavel-reply: I'm not convinced this whole test function is sensible or useful. Why do we need such behavior of this plugin?

plugins/cmd/tech_radar/main_test.go
```diff
@@ -0,0 +1,170 @@
...
+func TestCollectFailsOnInvalidConfig(t *testing.T) {
+	path := writeConfig(t, "schema_version: 1\nradar:\n  bogus: field\nquadrants: []\nrings: []\nentries: []\n")
+	p := newTestPlugin(t, path, testLogger())
+
+	response, err := p.Collect(t.Context())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), path)
+	assert.Contains(t, err.Error(), `unknown field "bogus" in radar`)
+	assert.Contains(t, err.Error(), "line 3")
+	assert.Empty(t, response.Nodes)
```
akavel-reply: Are we sure we need this assertion?

plugins/cmd/tech_radar/main_test.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func TestCollectRereadsFileEveryRun(t *testing.T) {
+	path := writeConfig(t, sampleConfig(t))
+	p := newTestPlugin(t, path, testLogger())
+
+	first, err := p.Collect(t.Context())
+	require.NoError(t, err)
+	require.Len(t, nodesByKind(first, pluginapi.NodeKindTechRadarEntry), 2)
+
```
akavel-reply, line 131: It'd be good to verify values before the change. Currently, I might be able to craft the sample-config.yaml such that it'd still pass this test even if the "...Rereads..." behavior is not working.

plugins/cmd/tech_radar/config.go
akavel-reply: This doesn't look fixed, the function still has a "naked" `return err` line.

Instead of passing `"radar config"` into `decodeStrict()`, please change this line to e.g.:

```suggestion
		return fmt.Errorf("in radar config: %w", err)
```

plugins/cmd/tech_radar/config.go
akavel-reply: FWIW, the problem I had here is now mostly solved with the Readme being more prominent in the plugin's directory.

akavel-reply:
> > > Thinking about, that this doesn't necessarly make sense as a standalone plugin with yet another container. I could keep the tech radar logically a plugin but run it in-process inside the catalog binary instead of as a gRPC sidecar.
> > 
> We could talk about it @akavel-reply. For me, it's even a question if this should be a plugin. (...)
> 
> The first option for me would be just to make this a standalone UI plugin, without backend. Parsing this config file for the tech-radar could be done in TypeScript. Pros: The tech-radar really becomes a plugin that can be installed and removed. Cons: There is no backend, so changing the config would require a rebuild of the UI-plugin. There will be no "last known good", without backend
> 
> Second option. Maybe if it's tightly coupled in the UI, we could also couple it tightly in the backend, and then just create a separate `techradar` package in the `catalog/internal` directory. Instead of trying to pass the tech-radar config through nodes and relations, we could just create new endpoints with a suitable API structure and mount them in `catalog/internal/httpapi/router.go`

Regarding the second option: if we want to move it outside of plugins, I would prefer not adding a custom new API endpoint to catalog, but then going all the way, and just serving it statically from some nginx on the side. Adding a new custom endpoints adds to maintenance and complexity burden. On the other hand, from technical and maintenance point of view, personally I'm ok with this being plugin. Maybe it's a very basic one, but I can imagine a user abusing (?) the catalog for something like this, so it's a somewhat interesting corner case. Technically, I really like that I can completely ignore existence of this plugin when thinking about the design of the catalog backend - _it's just another plugin_, and from my perspective, smoothly goes through the existing mechanisms.

As to the frontend - I do have some thoughts, but I decided I personally will limit myself to focusing on the backend & API in Naira, and leave the responsibility for the Naira UI to someone else, someone better qualified. I'm not very good in UI technologies, and there's more than enough work for me to stay on top of in the backend already. I'm happy to share or discuss my personal loose thoughts on the UI aspects in a loose offline discussions. (In shortest words - personally, I think we might want to have some kind of UI plugins, and the UI part of the tech_radar should then be such an UI plugin. But without having good knowledge of UI tech stacks, I can't easily design how such plugins system should work, and I don't see myself as a best person for that. Notably, it's also very unclear to me how we could make such plugins system in the UI secure to any extent.)

plugins/cmd/tech_radar/config.go
mkorbi: It could safe us some space to replace it but it would also break some logic I think. The report(...) calls in validate() (config.go:221-327) anchors its message to q.line / r.line / e.line. 
So, without the custom unmarshalers those are all zero, so "duplicate ring id", "references undeclared quadrant", "field must not be empty" etc. would all say line 0.

plugins/cmd/tech_radar/sample_config.yaml
mkorbi: indeed, changed

plugins/cmd/tech_radar/main_test.go
mkorbi: correct

plugins/cmd/tech_radar/main_test.go
mkorbi: name is maybe not the best, renamed it to `TestCollectFailsWithoutTouchingResponseOnMissingFile` to make it more clear. 

the assert.Empty I removed

plugins/cmd/tech_radar/main_test.go
mkorbi: deleted

plugins/cmd/tech_radar/main_test.go
mkorbi: indeed, fix is coming

plugins/cmd/tech_radar/main.go
mkorbi: understood, just merged the old readme and the godoc outcome, should be fixed

plugins/cmd/tech_radar/config.go
```diff
@@ -83,7 +83,7 @@ func (c *radarConfig) UnmarshalYAML(node *yaml.Node) error {
 	type alias radarConfig
 	decoded, line, err := decodeStrict[alias](node, "radar config")
```
akavel-reply: Can we now remove the string argument? I see it's passed into `checkKnownFields()`, but maybe it's not needed there also when the wrapping message now explains the parent field name?

plugins/cmd/tech_radar/main_test.go
```diff
@@ -121,7 +120,13 @@ func TestCollectRereadsFileEveryRun(t *testing.T) {
...
 	first, err := p.Collect(t.Context())
 	require.NoError(t, err)
-	require.Len(t, nodesByKind(first, pluginapi.NodeKindTechRadarEntry), 2)
+	// Pin the pre-change values: if the sample config drifted to already carry
+	// the post-change values, the assertions below would pass without any
+	// re-reading happening.
+	require.Equal(t, "2026-09", nodesByKind(first, pluginapi.NodeKindTechRadar)["naira"]["edition"])
+	firstEntry := nodesByKind(first, pluginapi.NodeKindTechRadarEntry)["naira/claude-sonnet"]
+	require.NotNil(t, firstEntry)
+	require.Equal(t, "in", firstEntry["moved"])
```
akavel-reply: (nitpick) It would be nice if the shape of the code here was matching the shape of the code in the block below - they both are doing the same thing, but for some reason they look differently, which makes them more tiresome to compare for a human reader, and thus makes spotting any mistakes harder by obscuring them. (FWIW, that's one of many great hiding techniques someone could try to use as part of their effort if they wanted to smuggle some security hole into a codebase.)

plugins/cmd/tech_radar/README.md
```diff
@@ -1,65 +1,12 @@
...
-
-In-portal editing, historic editions, enforcement of radar decisions, multiple radars per instance, and Git-based config sync are all out of scope for now - the mounted-file mechanism composes with any external GitOps tooling.
+The file is re-read on every collect, so ConfigMap updates take effect on the next sync without a redeployment. An invalid file fails the collect with errors naming the offending line and field, which should keep the previous radar snapshot in place (the catalog should never apply a failed run).
 
 ## Environment Variables
 
   - TECH\_RADAR\_CONFIG\_PATH (optional) - path to the radar YAML file; defaults to /etc/naira/techradar/radar.yaml.
+
+---
+Readme created from Go doc with [goreadme](https://github.com/posener/goreadme)
```
akavel-reply, line 12: _If_ your final decision is actually to generate the readme indeed by running goreadme, is there a reason you still didn't want to add a `//go:generate ...` line into `main.go`?

plugins/cmd/tech_radar/sample_config.yaml
```diff
@@ -1,9 +1,8 @@
 # Sample tech radar configuration.
 #
-# Schema reference: plugins/cmd/tech_radar/README.md. The plugin reads the config from
-# /etc/naira/techradar/radar.yaml by default; TECH_RADAR_CONFIG_PATH overrides
-# the path. This file is parsed by the plugin's unit tests, so it is always a
-# valid, up-to-date example.
+# Schema reference and usage details: plugins/cmd/tech_radar/README.md.
```
akavel-reply: With a re-run of goreadme, that you seem to have committed, we now lost this schema reference in the README.md file it appears, no? FWIW, I think keeping an explanation of the schema just in one place, right here within the `sample_config.yaml`, might be a cool idea, but even if that's indeed your decision, the text on this line now looks invalid and misleading.

plugins/cmd/tech_radar/main_test.go
akavel-reply: I gave a change suggestion, and you said "correct", but it looks like you didn't actually apply it?

mkorbi: seems like I missed to add the changes in main_test, comes with the next PR ^^

plugins/cmd/tech_radar/main_test.go
mkorbi: ok

akavel-reply (approved):
One comment seems still not addressed: https://github.com/naira-project/naira/pull/175#discussion_r3959194041 but generally I can approve.

plugins/cmd/tech_radar/config.go
mkorbi: fixed

mkorbi:
@akavel-reply I skipped the one comment, but fixed it now

hosseinsalahi (approved):

PR #167, opened 2026-08-27
feat(ui): lint UI files with biome

libera13:
## Description

This PR fixes all errors raised by `biome lint` 

Now
Screenshot 2026-08-27 at 13 31 00

before
Screenshot 2026-08-27 at 13 33 58

Note: I switched off noUnknownAtRules for css files. As biome doesn't recognize tailwind rules 
"Unknown at rule @tailwindcss(unknownAtRules)"

## Type of Change
- [x] 🔧 Refactoring (no functional changes)

PR #176, opened 2026-09-02
ci(ui): run biome and tsc checks in PR validation

### LINKED ISSUE #170: CI: biome and ts lint should run in CI
The `npm run check` command should be run in CI in the PR-validation workflow for UI.

Benefits: It runs `biome check` (which checks formatting and linting) and `tsc -b` (for type checking).

Here are the changes that work, although I haven't checked the overall quality of this workflow:
https://github.com/naira-project/naira/commit/1db2bdcd8086a90d1d165d5894d46177f3d96b1c

Here is the PR where it was tested to verify that it works:
https://github.com/naira-project/naira/pull/169

jagdeepm:
## Description

Adds a `ui-check` job to `pr-validation.yml` that installs UI dependencies and runs `tsc -b` + `biome ci` (via the existing `npm run check` logic), wired into the `all-checks` gate so a lint or type-check failure blocks merge like `go-test` already does.

Biome runs with `--reporter=github`, so failures show up as inline annotations directly on the PR diff instead of only in the raw job log:

Screenshot 2026-09-02 at 16 20 42

One wrinkle: biome reports file paths relative to the job's `working-directory: ui`, but GitHub resolves annotation `file=` paths relative to the repo root — so unprefixed paths silently failed to line up with the diff. The `Run checks` step pipes biome's output through `sed` to prefix `ui/` onto each path (with `set -o pipefail` so a lint failure still fails the step).

Builds on the CI wiring already prototyped in #169 — the code-level lint/type fixes from that PR have since landed independently via #167, so this PR only adds the missing CI job itself, plus the GitHub-annotation reporter and path fix.

## Type of Change

- [ ] 📝 Documentation update

## How Has This Been Tested?

- Verified directly against this PR's own CI runs: a deliberate `debugger` statement temporarily added to `Overview.tsx` caused `Check · UI` to fail with a correctly-placed inline annotation on the PR diff (screenshot above); removing it restored a green run.
- `cd ui && npm ci && npm run check` passes locally.

## Checklist

- [ ] I have added tests that prove my fix/feature works — N/A, this is a CI workflow change; verified live against this PR's own runs (see above)

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -70,8 +70,12 @@ jobs:
 
       - name: Run checks
         run: |
+          set -o pipefail
           npx tsc -b
-          npx biome ci --reporter=github .
+          # biome reports paths relative to this job's working-directory (ui/), but
+          # GitHub Actions resolves annotation `file=` paths relative to the repo
+          # root, so prefix them to make inline PR annotations line up correctly.
+          npx biome ci --reporter=github . | sed -E 's/(file=)/\1ui\//'
 
   # ── 4. Container build — catalog (build-only, no push on PRs) ─────────────
   container-build-catalog:
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -47,7 +47,31 @@ jobs:
       # revisit once the test suite reaches meaningful coverage.
       coverage-threshold: 0
 
-  # ── 3. Container build — catalog (build-only, no push on PRs) ─────────────
+  # ── 3. UI checks (TypeScript and Biome) ───────────────────────────────────
+  ui-check:
+    name: Check · UI
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ui
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          cache: npm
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-validation.yml
```diff
@@ -47,7 +47,37 @@ jobs:
...
+  # ── 3. UI checks (TypeScript and Biome) ───────────────────────────────────
+  ui-check:
+    name: Check · UI
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ui
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
libera13: Shouldn't we use actions/checkout@v7?

.github/workflows/pr-validation.yml
```diff
@@ -47,7 +47,37 @@ jobs:
...
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
```
libera13: I see actions/setup-node@v7 exists too

libera13 (approved):

libera13 (commented):
miss click before :)

.github/workflows/pr-validation.yml
jagdeepm: updated the version

.github/workflows/pr-validation.yml
jagdeepm: updated the action version

libera13 (approved):

kocakkaan (approved):

Issue #30, opened 2026-06-02
Inference Endpoints Catalog


mkorbi:
## 📋 Overview

User Story: As AI Engineer rAIner, I want to see which inference endpoints are actively running and serving requests, which models they expose, and through which provider or runtime, so that I understand where and how models are actually invocable.

User Story: As Platform Peter, I want all actively serving inference endpoints, whether routed through LiteLLM, hosted via vLLM/KServe, or available through external providers like OpenAI or Anthropic, to appear in the Naira catalog with their operational status, model links, and provider metadata, so that I can govern AI access across the organization.

User Story: As App Dev Abdel, I want to discover approved, actively running inference endpoints with clear status indicators and links to the models they serve, so that I know which endpoints I can integrate into my application right now.

The Inference Endpoints Catalog makes actively running inference interfaces visible as first-class entities in Naira's context engine. It does not deploy or manage endpoints; it discovers, catalogs, and contextualizes them.

---

## 🎯 Context and Problem Statement

The Model Catalog answers "what models exist?" The AI API Gateway integration answers "how is traffic routed?" But neither answers the most immediate operational question:

Which inference endpoints are actually running right now, and which models do they serve?

Today, the LiteLLM plugin (plugins/cmd/litellm) already queries the OpenAI-compatible /v1/models endpoint and creates model nodes in the catalog. But it does not create inference endpoint entities. The models appear in the catalog, but there's no representation of where those models are served from, their operational status, or the runtime behind them.

This means:

A user sees a model in the catalog but doesn't know if it's actually available for inference right now.
There's no way to distinguish between a model registered in MLflow (a training artifact) and a model actively served via LiteLLM or vLLM.
Multiple endpoints may serve the same model (e.g., LiteLLM route + direct vLLM endpoint), this topology is invisible.
Application developers have no catalog-backed way to discover which endpoints they can call.

### What already exists
 
| Component | Status | What it does |
|---|---|---|
| `catalog/internal/catalog/` | Implemented | In-memory graph store with Node, Relation, PluginClaim model |
| `plugins/pkg/pluginapi/` | Implemented | Plugin interface (`Collect` → `CollectResponse` with NodeClaim + RelationClaim), gRPC transport |
| `plugins/pkg/pluginapi/schema.go` | Implemented | Defines `NodeKindModel`, `NodeKindApplication`, `NodeKindDeployment`, `NodeKindService` etc. — but no `NodeKindInferenceEndpoint` |
[...]

PR #182, opened 2026-09-10
feat: inference endpoints catalog

### LINKED ISSUE #145: Schema Extension + LiteLLM Plugin Enhancement for IE
# Overview

As a Naira developer, I want inference_endpoint to be a first-class node kind in the catalog schema and the LiteLLM plugin to emit inference endpoint nodes alongside model nodes, so that actively served endpoints are discoverable through the existing catalog API.

# Scope

This task covers three changes:

- Schema constants — Add NodeKindInferenceEndpoint, RelationKindServesModel, and RelationKindExposedByGateway to plugins/pkg/pluginapi/schema.go.
- LiteLLM plugin enhancement — Extend the Collect() method in plugins/cmd/litellm/main.go to emit inference_endpoint nodes and serves_model relations for every model returned by /v1/models.
- Richer metadata via /model/info — Add a new HTTP call to LiteLLM's /model/info endpoint to retrieve upstream provider, api_base, model mode, and other LiteLLM-specific metadata. Populate endpoint properties from this response.

### LINKED ISSUE #146: Inference Endpoints UI Views
# Overview

As AI Engineer rAIner, I want to browse all actively serving inference endpoints in the Naira dashboard, filter by provider and status, and click through to the model each endpoint serves, so that I can find the right endpoint for my application.

As App Dev Abdel, I want to see which inference endpoints serve a specific model when I look at the model detail page, so that I know where to point my application for inference.

# Scope

Build three UI components in the OpenMFP dashboard:

1. Inference Endpoints List View (ideas)

A new micro-frontend page at /inference-endpoints in the OpenMFP navigation.

Columns: Endpoint name (path), Status (badge), Provider, Model served, Endpoint type, API protocol, Last seen.

Filters: Provider (multi-select), Status (multi-select), Endpoint type (multi-select).

Status badges:

active → green
degraded → yellow
unavailable → red
unknown → gray

Endpoint type labels:

gateway-route → "Gateway Route"
internal-hosted → "Self-Hosted"
external-managed → "External (Managed)"
external-self-hosted → "External (Self-Hosted)"

2. Inference Endpoint Detail View

Accessible by clicking an endpoint in the list view. Shows:

All endpoint properties in a structured layout.
Linked model(s) as clickable chips → navigates to model detail page.
Consuming applications (via reverse lookup: apps that uses_model the model this endpoint serves_model).
Source provenance: which plugin discovered this endpoint, last sync timestamp.

3. Model Detail Enhancement: "Served by" Section

On the existing model detail page add a "Served by" section that:

Queries GET /v1/relations?filter=kind="serves_model" filtered to relations where To matches the current model's NodeID.
Lists all inference endpoints serving this model with status badge and provider.
Each entry is clickable → navigates to the endpoint detail view.

If no inference endpoints serve the model, show a muted message: "No active inference endpoints discovered for this model."

### LINKED ISSUE #147: Additional Provider Plugins for IE
# Overview

As Platform Peter, I want Naira to discover inference endpoints from external providers like OpenAI or Anthropic, so that the Inference Endpoints Catalog reflects the full landscape of AI serving infrastructure available to my teams.

# Scope

This task creates additional provider plugins using the OpenAI-compatible helper that currently is implemented and the standard pluginapi.Plugin contract. Each plugin is a separate binary in plugins/cmd/.

## External Managed Providers

These require thin adapter logic wrapping provider-native APIs.

- OpenAI 
- Scaleway
- StackIT
- Anthropic
- AWS Bedrock
- Azure OpenAI
- Google Vertex AI

kocakkaan:
## Description

This PR is created mainly for the issue #30 and its subissues and aims to create an inference endpoint catalog within Naira. In the first part of PR, LiteLLM is used for creating actively running inference endpoint nodes. For mapping multiple inference endpoints into one model, `modelName` parameter in LiteLLM has been used. Inference endpoints' catalog gives the user available inference endpoints with detailed information fetched from LiteLLM's `/model/info` endpoint. #145 

In UI, main focus is on retrieving the important properties through GenericTable. Additional OpenMFP navigation tab has been added under `serviceProviders.ts`. `MCPServerTools.tsx` and `detailTabs.ts` files have been merged under `detailTabs.tsx` for providing relationships between different kinds. #146 

Additionally, AWS Bedrock plugin is added on how its inference endpoints can be consumed. The codebase lists foundational models and their inference endpoints for a specific IAM user. Once an inference endpoint is invoked in AWS Bedrock, CloudWatch fetches the metrics of this specific instance and they can be observed in Naira as additional properties. #147

## Type of Change

- [x] 📝 Documentation update

## How Has This Been Tested?

- [ ] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

## AI Usage
Sonnet 5 is used for coding.


ui/src/config/viewpoints.ts
```diff
@@ -47,6 +47,13 @@ export const CATALOG_VIEWPOINTS: CatalogViewpoint[] = [
     columns: ['endpoint', 'description', 'instructions'],
     plugins: ['mcp-servers', 'litellm'],
   },
+  {
+    path: 'inference_endpoints',
+    heading: 'Inference Endpoints Catalog',
+    subheading: 'Inference endpoints served via different plugins.',
+    kinds: ['inference_endpoint'],
+    plugins: ['litellm', 'bedrock'],
```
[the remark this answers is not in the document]
kocakkaan: Needed to be done later with custom pages.

plugins/pkg/pluginapi/schema.go
```diff
@@ -24,4 +25,6 @@ const (
 	RelationKindDeployedFrom = "deployed_from"
 	RelationKindDerivedFrom  = "derived_from"
 	RelationKindExposes      = "exposes"
+	RelationKindServesModel  = "serves_model"
+	RelationKindExposedByGateway = "exposed_by_gateway"
```
akavel-reply, line 29: Please make sure to use `gofmt` to format your Go files. (Note: this can be typically configured in your editor's Go plugin.)

plugins/cmd/litellm/main.go
```diff
@@ -69,6 +69,7 @@ func (p *Plugin) Collect(ctx context.Context) (pluginapi.CollectResponse, error)
 	seenRelations := make(map[string]struct{})
 	collectErrors := make([]error, 0)
 
+	ownedByModel := make(map[string]string, len(models))
```
akavel-reply: please rename `ownedByForModel` or `ownedByPerModel`, or otherwise something more clear. Currently, "owned by model" suggests that a model owns something, and also doesn't clarify what this model owns. From looking how it's used, this understanding seems incorrect, so the name should be changed to describe correct semantics/meaning.

plugins/cmd/litellm/main.go
```diff
@@ -69,6 +70,7 @@ func (p *Plugin) Collect(ctx context.Context) (pluginapi.CollectResponse, error)
...
```
akavel-reply: please rename `ownedByForModel` or `ownedByPerModel`, or otherwise something more clear. Currently, "owned by model" suggests that a model owns something, and also doesn't clarify what this model owns. From looking how it's used, this understanding seems incorrect, so the name should be changed to describe correct semantics/meaning.

Also, please rename the same way the argument name in function `listInferenceEndpoints()`.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, fmt.Errorf("litellm /model/info returned %s", resp.Status)
+	}
+
+	var payload modelInfoResponse
+	if err := json.NewDecoder(resp.Body).Decode(&payload); err != nil {
+		return nil, fmt.Errorf("decoding LiteLLM Model info response: %w", err)
+	}
```
akavel-reply: Most of this is practically identical in `fetchEndpointsHealth()`, and probably also more functions (see e.g. all places where `addAuthorization()` is called). Please look into extracting some common helper function - at least for the most similar ones (`fetchAllowedModels()` seems a bit more complex, you can leave it aside initially; but maybe the code building the `.RawQuery` field could be moved before and use just regular URL argument, and it could still then work same as others).

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+
+	statusByKey, err := p.fetchEndpointHealth(ctx)
```
akavel-reply: I think we should move this call outside of `fetchInferenceEndpoints()`, and pass the result as argument into this function if it's needed. Currently, there's some mixing of abstraction levels which makes things harder to read.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	return status, nil
+}
+
+// It is used for mapping status into respective endpoints.
+// There are three status states now; healthy, unhealthy, and unknown.
+func endpointHealthKey(model, apiBase string) string {
+	return strings.TrimSpace(model) + "|" + strings.TrimSpace(apiBase)
+}
```
akavel-reply: Instead of concatenating strings, please remove this func and use an explicit struct instead - Go allows structs as keys in a map. For example:

```go
type modelAndAPIBase struct { model, apiBase string }

foobar := map[modelAndAPIBase]string{}
foobar[modelAndAPIBase{"foo", "bar"}] = "foobar"
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("calling LiteLLM Model info endpoint: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, fmt.Errorf("litellm /model/info returned %s", resp.Status)
+	}
+
+	var payload modelInfoResponse
```
akavel-reply: I think we could reduce the number of types and their names a bit, by refactoring this to:

```suggestion
	var payload struct {
		Data []struct{
			ModelName     string        `json:"model_name"`
			LiteLLMParams liteLLMParams `json:"litellm_params"`
			ModelInfo     modelInfo     `json:"model_info"`
		} `json:"data"`
	}
```

akavel-reply (commented):
First bunch of comments; I haven't done full round of review yet.

Caveat: I will only review the backend parts of the PR, with focus on the Go parts.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("listing inference endpoints: %w", err)
+	}
+
+	invocations, err := p.fetchModelInvocations(ctx)
+	if err != nil {
+		if p.logger != nil {
+			p.logger.Printf("WARN: fetching LiteLLM daily activity failed, continuing without invocation counts: %v", err)
+		}
+		invocations = map[string]int64{}
```
akavel-reply, line 138: It looks like in this case, the following loop never builds any nodes or relations; is this on purpose? if yes, we should instead just return early here with empty outputs.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+
+		endpoint.OwnedBy = ownedByModel[modelName]
+
+		endpointKey := modelName
+		if region := strings.TrimSpace(endpoint.Region); region != "" {
+			endpointKey += "-" + region
+		}
+
+		endpointNode := pluginapi.NodeClaim{
+			ID:         pluginapi.NodeID{Kind: pluginapi.NodeKindInferenceEndpoint, Path: p.config.PathPrefix + "/" + endpointKey},
```
akavel-reply, line 170: please split this to multiline for easier readability:

```suggestion
			ID:         pluginapi.NodeID{
				Kind: pluginapi.NodeKindInferenceEndpoint,
				Path: p.config.PathPrefix + "/" + endpointKey,
			},
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+			continue
+		}
+		endpoint.Invocations = invocations[modelName]
+
+		endpoint.OwnedBy = ownedByModel[modelName]
+
+		endpointKey := modelName
+		if region := strings.TrimSpace(endpoint.Region); region != "" {
+			endpointKey += "-" + region
+		}
```
akavel-reply, line 167: Please tweak as below and adjust Path expression accordingly - this will make it slightly easier to see on the spot how the Path expression is built:

```suggestion
		regionSuffix := ""
		if region := strings.TrimSpace(endpoint.Region); region != "" {
			regionSuffix = "-" + region
		}
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+		endpointNode := pluginapi.NodeClaim{
+			ID:         pluginapi.NodeID{Kind: pluginapi.NodeKindInferenceEndpoint, Path: p.config.PathPrefix + "/" + endpointKey},
+			Properties: endpoint.properties(),
+		}
+		nodes = append(nodes, endpointNode)
+
+		relations = append(relations, pluginapi.RelationClaim{
+			Kind: pluginapi.RelationKindServesModel,
+			From: endpointNode.ID,
+			To:   pluginapi.NodeID{Kind: pluginapi.NodeKindModel, Path: p.config.PathPrefix + "/" + modelName},
```
akavel-reply, line 178: please split this to multiline struct assignment for easier readability (see a comment above for an example)

plugins/cmd/litellm/inference_endpoints_test.go
```diff
@@ -0,0 +1,195 @@
...
+	mux.HandleFunc("/health", func(w http.ResponseWriter, _ *http.Request) {
+		if health == nil {
+			http.Error(w, "unavailable", http.StatusServiceUnavailable)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		require.NoError(t, json.NewEncoder(w).Encode(health))
+	})
+	mux.HandleFunc("/user/daily/activity", func(w http.ResponseWriter, _ *http.Request) {
+		modelGroups := make(map[string]dailyActivityModelEntry, len(invocationsByModel))
```
akavel-reply, line 35: If possible, in test inputs and outputs please try to use explicit, raw strings, rather than "proper" structs etc. This is on purpose opposite to "normal" good coding practices: in a test, it is preferable to see explicit, "dumb" input/output, so looking at it it's easier to see what is really being processed.

So, here, please if possible return just a "dumb", raw JSON string - something like:
```go
w.Write([]byte(`
{
  "results": [{
    "breakdown": {
      "model_groups": ...
    }
  }]
}
`))
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+			ModelID:      entry.ModelInfo.ModelID,
+			Provider:     entry.LiteLLMParams.provider(),
+			EndpointType: entry.LiteLLMParams.endpointType(),
+			Status:       status,
+			EndpointURL:  entry.LiteLLMParams.APIBase,
+			Region:       entry.LiteLLMParams.RegionName,
+			ServesModel:  strings.TrimSpace(entry.ModelName),
+			Mode:         entry.ModelInfo.Mode,
+			MaxTokens:    entry.ModelInfo.MaxTokens,
+			InputCost:    entry.ModelInfo.InputCostPerToken,
```
akavel-reply, line 307: Please prefer to avoid renaming fields from the names used by the underlying services. If we show a raw value copied from a service, let's keep the same name that was used there. This will make it much easier for our end-user to find explanation of what that value means by looking into the documentation of the specific underlying service's API (in this case, LiteLLM API documentation).

If there's some notable reason why we _do_ want to rename a field, please give a comment with a rationale in the code, in place where this renaming visibly happens (i.e. it would be here).

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+// fetchModelInvocations queries LiteLLM's /user/daily/activity for the
+// configured lookback window, without a user_id filter, so the master key
+// gets api_requests summed across all users for each model.
+// TODO: configure fetch mechanism to be scoped to a specific id, however it is to be talked
+// together with the authorization mechanism within Naira (i.e. how to map LiteLLM credentials with Naira user).
+func (p *Plugin) fetchModelInvocations(ctx context.Context) (map[string]int64, error) {
+	lookback := p.config.MetricsLookback
+	if lookback <= 0 {
+		lookback = 24 * time.Hour
+	}
```
akavel-reply, line 350: We shouldn't need this - we already specified a default value in the config field, this behavior is quite surprising, and also seems undocumented.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,386 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("calling LiteLLM daily activity endpoint: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, fmt.Errorf("litellm /user/daily/activity returned %s", resp.Status)
+	}
+
+	var payload dailyActivityResponse
```
akavel-reply, line 363: Please inline the type definition here (from what I can see, I think it should be possible if we return raw JSON strings in tests) - example:

https://github.com/naira-project/naira/blob/14f78f8d9d0bb15a76fd48f6eea89ec780ee787f/plugins/cmd/github/attestation.go#L29-L35

If you still want to split it a bit because it becomes too nested/complex, please define all needed types here locally, just above this line (types can be defined inside functions in Go).

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,395 @@
+package main
```
akavel-reply, line 1: We try to document a plugin in a godoc comment, and generate a README.md file from that godoc, using a `go generate` command with `goreadme` tool fetched via `mise`. Could you try to convert the readme of this plugin into such godoc comment, and run the `go generate` command with it?

See for example (note especially the `//go:generate ...` line at the end):
https://github.com/naira-project/naira/blob/14f78f8d9d0bb15a76fd48f6eea89ec780ee787f/plugins/cmd/depl_uses_litellm/plugin.go#L1-L36

plugins/cmd/bedrock/README.md
```diff
@@ -0,0 +1,27 @@
...
+Currently, AWS Bedrock model fetches `all` foundational models and their inference endpoints available for a specific IAM user. However, this fetch now results with ~140 available inference endpoints, if the region is selected as `us-east-1` and ~40-50 available inference endpoints, if the region is selected as `eu-central-1`. This fetch mechanism provides small delay on the fetch, but with more regions available for a specific IAM user, this mechanism must be optimized. 
+
+
+## Environment Variables
+
+  - `AWS_ACCESS_KEY_ID` (mandatory) - Access key ID of a specific IAM user instance. This key ID, alongside with this IAM user's secret access key, is used for authorizing user to consume resources that they are permitted to.
+  
+  - `AWS_SECRET_ACCESS_KEY` (mandatory) - Access key secret of a specific IAM user instance. This is used with access key ID as an authorization mechanism. 
+
+  - `BEDROCK_REGIONS` (optional) - Default region is specified as `us-east-1`.
```
akavel-reply, line 25: "REGIONS" means many can be specified, yes? if so, please explain how to specify multiple (should they be comma-separated? space-separated? otherwise?)

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,395 @@
...
+	config              config
+	newBedrockClient    func(ctx context.Context, region string) (bedrockClient, error)
+	newCloudWatchClient func(ctx context.Context, region string) (cloudWatchClient, error)
+}
+
+// bedrockClient is the subset of *bedrock.Client used by this plugin, so tests
+// can substitute a fake implementation.
+type bedrockClient interface {
+	ListFoundationModels(ctx context.Context, params *bedrock.ListFoundationModelsInput, optFns ...func(*bedrock.Options)) (*bedrock.ListFoundationModelsOutput, error)
+}
```
akavel-reply, line 69: If we only need one function, could we instead just define (and return) a closure (a ["function type"](https://go.dev/ref/spec#Function_types)), instead of a whole interface? this should simplify this code, possibly remove need for that comment, and might also make mocking easier in tests.

It might then also be possible to drop arg names from the function type definition - the types seem self-explanatory, the arg names don't seem to really add anything, and this could possibly make the function signature slightly shorter.

Same also for the interface below.

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,395 @@
...
+		properties[propertyKeyEndpointStatus] = usage.status()
+	}
+
+	return properties
+}
+
+func (p *Plugin) listFoundationModels(ctx context.Context, region string) ([]foundationModel, error) {
+	client, err := p.newBedrockClient(ctx, region)
+	if err != nil {
+		return nil, err
```
akavel-reply, line 250: please don't return "naked errors", always wrap error in `fmt.Errorf` (see our AGENTS.md) - or if not, add a comment explaining why we're doing an exception in given place.

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,395 @@
...
+
+	models := make([]foundationModel, 0, len(out.ModelSummaries))
+	for _, summary := range out.ModelSummaries {
+		models = append(models, foundationModel{
+			ModelID:          derefString(summary.ModelId),
+			ModelName:        derefString(summary.ModelName),
+			ProviderName:     derefString(summary.ProviderName),
+			LifecycleStatus:  strings.ToLower(string(lifecycleStatus(summary.ModelLifecycle))),
+			InputModalities:  modalitiesToStrings(summary.InputModalities),
+			OutputModalities: modalitiesToStrings(summary.OutputModalities),
```
akavel-reply, line 267: We seem to do some conversions here, then another conversions in `foundationModel.properties()`; could we just make all conversions in one place? or at least as many as possible?

Could we just keep already fully converted props in a `foundationModel` struct? if not, if we need some fields for some other uses elsewhere in the code (I see e.g. `ModelName` seems used like so at a first glance), maybe we could keep only those fields outside, but still also copy them into props if needed? e.g.:
```go
foundationModel{
    ModelID: derefString(summary.ModelId),
    Props: map[string]string{
        propertyKeyModelID: derefString(summary.ModelId), // if needed
        propertyKeyProviderName: derefString(summary.ProviderName),
        ...etc...
    },
}
```

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,395 @@
...
+// configured lookback window, so usage and health can be compared across
+// regions and models.
+func (p *Plugin) fetchTokenUsage(ctx context.Context, region string, models []foundationModel) (map[string]modelUsage, error) {
+	if len(models) == 0 {
+		return map[string]modelUsage{}, nil
+	}
+
+	client, err := p.newCloudWatchClient(ctx, region)
+	if err != nil {
+		return nil, err
```
akavel-reply, line 286: please make sure to wrap errors in fmt.Errorf

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	//TODO: use dynamic time intervals in the UI, not just from BEDROCK_METRICS_LOOKBACK
+	lookback := p.config.MetricsLookback
+	if lookback <= 0 {
+		lookback = metricLookbackWindow
+	}
```
akavel-reply, line 293: a default in config should be enough, we should need extra undocumented tweaks like this

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,395 @@
...
+
+func metricQuery(id, metricName string, dimensions []cwtypes.Dimension) cwtypes.MetricDataQuery {
+	namespace := metricNamespaceBedrock
+	stat := "Sum"
+	period := int32(metricPeriodSeconds)
+	return cwtypes.MetricDataQuery{
+		Id: &id,
+		MetricStat: &cwtypes.MetricStat{
+			Metric: &cwtypes.Metric{
+				Namespace:  &namespace,
```
akavel-reply, line 357: since Go ~1.26, it should be possible to write this without needing to define separate variable - please fix this and other similar cases in this func:

```suggestion
				Namespace:  new(metricNamespaceBedrock),
```

akavel-reply (commented):
Another set of comments.

plugins/cmd/litellm/inference_endpoints.go
kocakkaan: Yes, exactly. It is on purpose. I am changing the invocations part to empty node and relation claims. 👍

plugins/cmd/bedrock/README.md
kocakkaan: they should be space-separated, documenting it here.

plugins/cmd/litellm/inference_endpoints_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	httpServer := httptest.NewServer(mux)
+	t.Cleanup(httpServer.Close)
+
+	return httpServer.URL
+}
+
+func TestListInferenceEndpointsEmitsNodesAndRelations(t *testing.T) {
+	baseURL := startLiteLLMModelInfo(t,
+		modelInfoResponse{Data: []modelInfoEntry{
```
akavel-reply: can we pass raw JSON string here, instead of building a struct?

```go
const modelInfoResponse = `{
  "data": [
    {
      "model_name": "idp-claude-sonnet",
      ...
    }
  ]
}`
const healthResponse = `{
  "healthy_endpoints": [
    {
      "model": "anthropic/claude...",
      "api_base": "..."
  ]
}`
const dailyActivityResponse = `{
  ....
}`
baseURL := startLiteLLMModelInfo(t, modelInfoResponse, healthResponse, dailyActivityResponse)

File: plugins/pkg/pluginapi/schema.go [hunk 26]
```diff
@@ -11,17 +11,20 @@ const (
...
+	RelationKindTrainedOn        = "trained_on"
+	RelationKindUsesModel        = "uses_model"
+	RelationKindCalls            = "calls"
+	RelationKindSourcedFrom      = "sourced_from"
+	RelationKindDescribes        = "describes"
+	RelationKindDeployedFrom     = "deployed_from"
+	RelationKindDerivedFrom      = "derived_from"
+	RelationKindExposes          = "exposes"
+	RelationKindServesModel      = "serves_model"
+	RelationKindExposedByGateway = "exposed_by_gateway"
```
libera13: RelationKindExposedByGateway looks not used

ui/src/config/detailTabs.tsx
```diff
@@ -0,0 +1,58 @@
+import { BrainCircuit, Cloud, Wrench } from 'lucide-react';
+import type { RelatedNodesConfig } from '../components/RelatedNodes';
+
+export const MCP_SERVER_KIND = 'mcp_server';
+const RELATION_KIND_SERVES_MODEL = 'serves_model';
```
libera13: 1. MCP_SERVER_KIND export looks not needed
2. naming can be more consistent for example:
```suggestion
const MCP_SERVER_KIND = 'mcp_server';
const SERVES_MODEL_RELATION = 'serves_model';
```

ui/src/config/detailTabs.tsx
```diff
@@ -0,0 +1,58 @@
...
+  primary?: boolean;
+}
+
+export const RELATED_CARDS_BY_KIND: Record<string, RelatedNodesConfig> = {
+  mcp_server: {
+    relationKind: 'exposes',
+    direction: 'outgoing',
+    title: 'tools',
+    icon: Wrench,
+    countSuffix: 'exposed',
```
libera13: if it's not crucial to have this countSuffix, maybe remove it, then it will be smaller config

ui/src/components/RelatedNodes.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { useNavigate } from 'react-router';
+import { type CatalogGraphResponse, useCatalogGraph } from '../hooks/useCatalogGraph';
+import { encodeCatalogPath, type NodeResource } from '../lib/catalogApi';
+import { parsePath } from '../lib/kindUtils';
+
+/**
+ * Describes one set of neighbours reached from a node via a single relation
+ * kind/direction, e.g. the tools an MCP server exposes, or the model an
+ * inference endpoint serves. Drives both the Tools tab (`layout: 'list'`) and
+ * the Properties-tab cross-link cards (`layout: 'cards'`, the default).
```
libera13: I cannot find what layout we are talking here about

ui/src/config/detailTabs.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <span className="font-medium">reachable</span> property.
+      </>
+    ),
+  },
+  inference_endpoint: {
+    relationKind: RELATION_KIND_SERVES_MODEL,
+    direction: 'outgoing',
+    title: 'Uses Model',
+    icon: BrainCircuit,
+    description: 'The model this inference endpoint serves.',
```
libera13: is the emptyText only needed in mcp_servers? inference_endpoint and model views cannot be in that state?

ui/src/components/RelatedNodes.tsx
```diff
@@ -0,0 +1,162 @@
...
+  /** Heading (cards layout) or item-type label (list layout). */
+  title: string;
+  /** Icon shown next to each item in list layout. Defaults to a chevron. */
+  icon?: ComponentType<{ size?: number; className?: string }>;
+  /** Word appended after the item count in list layout, e.g. "exposed". */
+  countSuffix?: string;
+  /** Shown instead of nothing when list layout has no items. Cards layout renders nothing either way. */
+  emptyText?: ReactNode;
+
+  description?: string;
```
libera13: this one only doesn't have a comment, so raised a interest in me what is it. Looks like mcp specific property.
Is it needed to have in generic component, field only for one plugin?
Anyway would be good to update docs here

libera13 (commented):
I looked only at UI

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -241,54 +234,67 @@ func (e inferenceEndpoint) properties() pluginapi.PropertyMap {
...
+func (p *Plugin) fetchInferenceEndpoints(ctx context.Context, statusByKey map[modelAndAPIBase]string) ([]inferenceEndpoint, error) {
+	var payload struct {
+		Data []struct {
+			ModelName     string           `json:"model_name"`
+			LiteLLMParams modelInfoLiteLLM `json:"litellm_params"`
+			ModelInfo     modelInfoDetail  `json:"model_info"`
+		} `json:"data"`
+	}
+	if err := p.getLiteLLMJSON(ctx, p.config.BaseURL+"/model/info", "Model info", "/model/info", &payload); err != nil {
+		return nil, err
```
akavel-reply: Instead of passing `"Model info", "/model/info"` into `getLiteLLMJSON()`, please wrap `err` here with `fmt.Errorf` and include that information there.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -241,54 +234,67 @@ func (e inferenceEndpoint) properties() pluginapi.PropertyMap {
...
-			MaxTokens:    entry.ModelInfo.MaxTokens,
-			InputCost:    entry.ModelInfo.InputCostPerToken,
-			OutputCost:   entry.ModelInfo.OutputCostPerToken,
+			ModelID:            entry.ModelInfo.ModelID,
+			Provider:           entry.LiteLLMParams.provider(),
+			EndpointType:       entry.LiteLLMParams.endpointType(),
+			Status:             status,
+			EndpointURL:        entry.LiteLLMParams.APIBase,
+			Region:             entry.LiteLLMParams.RegionName,
+			ServesModel:        strings.TrimSpace(entry.ModelName),
```
akavel-reply: I asked to not rename fields without explanation, but it still looks like you rename them without explanation. Please fix.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -241,54 +234,67 @@ func (e inferenceEndpoint) properties() pluginapi.PropertyMap {
...
-		}
+	return nil
+}
+
+func (p *Plugin) fetchInferenceEndpoints(ctx context.Context, statusByKey map[modelAndAPIBase]string) ([]inferenceEndpoint, error) {
+	var payload struct {
+		Data []struct {
+			ModelName     string           `json:"model_name"`
+			LiteLLMParams modelInfoLiteLLM `json:"litellm_params"`
+			ModelInfo     modelInfoDetail  `json:"model_info"`
```
akavel-reply: I asked for rename of the struct types here, and they are still not renamed. Please fix.
- `modelInfoLiteLLM` -> `liteLLMParams` (or `litellmParams`)
- `modelInfoDetail` -> `modelInfo`

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	endpoints, err := p.fetchInferenceEndpoints(ctx, statusByKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("listing inference endpoints: %w", err)
+	}
+
+	invocations, err := p.fetchModelInvocations(ctx)
+	if err != nil {
+		return []pluginapi.NodeClaim{}, []pluginapi.RelationClaim{}, fmt.Errorf("Error while fetching model invocations: %v", err)
```
akavel-reply: would it work to just return nils here?

```suggestion
		return nil, nil, fmt.Errorf("Error while fetching model invocations: %v", err)
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -298,81 +304,61 @@ func (p *Plugin) fetchInferenceEndpoints(ctx context.Context) ([]inferenceEndpoi
...
+				ModelGroups map[string]struct {
+					Metrics struct {
+						APIRequests int64 `json:"api_requests"`
+					} `json:"metrics"`
+				} `json:"model_groups"`
+			} `json:"breakdown"`
+		} `json:"results"`
+	}
+	if err := p.getLiteLLMJSON(ctx, requestURL.String(), "daily activity", "/user/daily/activity", &payload); err != nil {
+		return nil, err
```
akavel-reply: as mentioned in other comment, do not return "naked errors", wrap them with context through fmt.Errorf (see AGENTS.md)

plugins/cmd/bedrock/main.go
```diff
@@ -1,3 +1,30 @@
+// `bedrock` scans available foundational models and their inference endpoints for a specific IAM user fetched from a running AWS Bedrock instance.
+//
+// ## Setup
```
akavel-reply: Unfortunately, godoc format is not exactly the same as Markdown. The file generated by goreadme now looks not good if you view it as rendered markdown. Please tweak this godoc to improve how the generated and then rendered README.md looks. Notably:

- headers
- numbered list

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,418 @@
...
+// 2. Generate an access key for that user (IAM -> Users -> Security credentials -> Create access key, "Local code" use case). Copy the secret immediately or download the .csv file which consists of the credentials.
+// 3. Provide `BEDROCK_AWS_ACCESS_KEY_ID` and `BEDROCK_AWS_SECRET_ACCESS_KEY` to the plugin:
+//   - Keep placeholder/empty values in the tracked manifest, and instead create the secret out-of-band, e.g. `kubectl create secret generic catalog-secrets --from-literal=BEDROCK_AWS_ACCESS_KEY_ID=... --from-literal=BEDROCK_AWS_SECRET_ACCESS_KEY=...`.
+//
+// 4. Verify Bedrock model access in the configured regions (Bedrock -> Model catalog, matching `BEDROCK_REGIONS`), and confirm you've invoked at least one model per region. CloudWatch only reports `InputTokenCount`/`OutputTokenCount`/`Invocations` for models that have received traffic.Otherwise the plugin just shows zero usage.
```
akavel-reply: typo

```suggestion
// 4. Verify Bedrock model access in the configured regions (Bedrock -> Model catalog, matching `BEDROCK_REGIONS`), and confirm you've invoked at least one model per region. CloudWatch only reports `InputTokenCount`/`OutputTokenCount`/`Invocations` for models that have received traffic. Otherwise the plugin just shows zero usage.
```

ui/src/components/RelatedNodes.tsx
kocakkaan: both MCPServerTools and my detail tab for inference endpoints were using different layouts. I have unified their layout but forgot to delete this comment. Deleting the comment now.

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,418 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("calling Bedrock ListFoundationModels: %w", err)
+	}
+
+	models := make([]foundationModel, 0, len(out.ModelSummaries))
+	for _, summary := range out.ModelSummaries {
+		props := pluginapi.PropertyMap{}
+		for key, value := range map[string]string{
+			propertyKeyModelName:        derefString(summary.ModelName),
+			propertyKeyLifecycleStatus:  strings.ToLower(string(lifecycleStatus(summary.ModelLifecycle))),
```
akavel-reply: ToLower and conversion to string should be inside `lifecycleStatus` helper.

plugins/cmd/bedrock/main.go
```diff
@@ -0,0 +1,418 @@
...
+			propertyKeyInputModalities:  strings.Join(modalitiesToStrings(summary.InputModalities), ","),
+			propertyKeyOutputModalities: strings.Join(modalitiesToStrings(summary.OutputModalities), ","),
```
akavel-reply: modalitiesToStrings appears only used here so I assume we can move strings.Join into that func.

After that, we can probably simplify the func to call ToLower on the joined string, and thus remove the loop.

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -40,23 +40,23 @@ const (
...
-	OutputCost      float64 `json:"output_cost_per_token"`
-	Invocations     int64   `json:"-"`
+	ModelID            string  `json:"model_id"`
+	EndpointType       string  `json:"endpoint_type"`
+	Provider           string  `json:"provider"`
+	Status             string  `json:"status"`
+	APIProtocol        string  `json:"api_protocol"`
+	EndpointURL        string  `json:"endpoint_url"`
+	Region             string  `json:"region"`
+	ServesModel        string  `json:"model_name"`
```
akavel-reply: this field name should not differ from what is in json

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,372 @@
...
+	OwnedBy            string  `json:"owned_by"`
+	LifecycleStatus    string  `json:"lifecycle_status"`
+	DiscoveredVia      string  `json:"discovered_via"`
+	LastSeen           string  `json:"last_seen"`
+	Mode               string  `json:"mode"`
+	MaxTokens          int64   `json:"max_tokens"`
+	InputCostPerToken  float64 `json:"input_cost_per_token"`
+	OutputCostPerToken float64 `json:"output_cost_per_token"`
+	Invocations        int64   `json:"-"`
```
akavel-reply: Please add empty line above this field, to show that it is different from others because it's not deserialized from json:

```suggestion

	Invocations int64 `json:"-"`
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,372 @@
...
+	HealthyEndpoints   []healthEndpointEntry `json:"healthy_endpoints"`
+	UnhealthyEndpoints []healthEndpointEntry `json:"unhealthy_endpoints"`
+}
+
+type healthEndpointEntry struct {
+	Model   string `json:"model"`
+	APIBase string `json:"api_base"`
+}
+
+type modelAndAPIBase struct{ model, apiBase string }
```
akavel-reply: Please refactor the block above as shown below:
```suggestion
type healthResponse struct {
	HealthyEndpoints   []modelAndAPIBase `json:"healthy_endpoints"`
	UnhealthyEndpoints []modelAndAPIBase `json:"unhealthy_endpoints"`
}

type modelAndAPIBase struct {
	Model   string `json:"model"`
	APIBase string `json:"api_base"`
}
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,372 @@
...
+	LiteLLMParams modelInfoLiteLLM `json:"litellm_params"`
+	ModelInfo     modelInfoDetail  `json:"model_info"`
+}
+
+type modelInfoLiteLLM struct {
+	Model             string `json:"model"`
+	APIBase           string `json:"api_base"`
+	CustomLLMProvider string `json:"custom_llm_provider"`
+	RegionName        string `json:"region_name"`
+}
```
akavel-reply: With the change described below (see comment near `type modelAndAPIBase`), you should be able to change this to:

```suggestion
type modelInfoLiteLLM struct {
	modelAndAPIBase
	CustomLLMProvider string `json:"custom_llm_provider"`
	RegionName        string `json:"region_name"`
}
```

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,372 @@
...
+	"time"
+
+	"github.com/naira-project/naira/plugins/pkg/pluginapi"
+)
+
+const (
+	propertyKeyModelID                    = "model_id"
+	propertyKeyEndpointType               = "endpoint_type"
+	propertyKeyProvider                   = "provider"
+	propertyKeyEndpointStatus             = "status"
```
akavel-reply: Again, const name doesn't match the value, please fix

plugins/cmd/litellm/inference_endpoints.go
```diff
@@ -0,0 +1,372 @@
...
+	propertyKeyEndpointURL                = "endpoint_url"
+	propertyKeyRegion                     = "region"
+	propertyKeyServesModel                = "serves_model"
+	propertyKeyLifecycleStatus            = "lifecycle_status"
+	propertyKeyLastSeen                   = "last_seen"
+	propertyKeyMode                       = "mode"
+	propertyKeyMaxTokens                  = "max_tokens"
+	propertyKeyInputCostPerMillionTokens  = "input_cost_per_million_tokens"
+	propertyKeyOutputCostPerMillionTokens = "output_cost_per_million_tokens"
+	propertyKeyInvocations                = "invocations_total"
```
akavel-reply: again, name vs. value mismatch

plugins/cmd/litellm/inference_endpoints_test.go
akavel-reply: This looks still not done. Please apply. This should allow us to remove the `modelInfoResponse` type definition and many others.

akavel-reply (changes requested):

PR #187, opened 2026-09-15
feat: aligns the operations API with the AIP-151

libera13:
## Description

This PR aligns the operations API with the AIP-151 operation resource [format](https://github.com/googleapis/googleapis/blob/1d4ae2021761d93a0dfcd532b52f0cd87c20a8e2/google/longrunning/operations.proto#L121). The PR also moves "operations" unit tests from `router_test.go` to `operations_test.go`, with some refactoring on the way to make tests more readable.

Changes on UI:
- UI is updated to consume the new response shape.
- Replace usage of raw html buttons with using Button from "components/ui". Thanks to it we have more consitent styling and behavior (e.g. cursor pointer on hover)
- Removed top bar from CatalogView. In topbar was "Search kinds" and "Last sync" info.

## Type of Change

- [ ] 📝 Documentation update
- [x] 🔧 Refactoring (no functional changes)

## How Has This Been Tested?

- [x] `go test ./...` passes
- [ ] `helm lint` passes

## Checklist

### DIFF
--- a/catalog/internal/httpapi/operations.go
+++ b/catalog/internal/httpapi/operations.go
@@ -18,17 +18,29 @@ type StatusErrorResource struct {
 	Message string `json:"message"`
 }
 
+// OperationMetadataResource holds progress information about an in-flight
+// or completed operation. It never carries the operation's result.
+type OperationMetadataResource struct {
+	Plugin    string     `json:"plugin"`
+	State     string     `json:"state"`
+	StartTime time.Time  `json:"startTime"`
+	EndTime   *time.Time `json:"endTime,omitempty"`
+	CreatedAt time.Time  `json:"createdAt"`
+}
+
+// RunPluginResult is the successful result of a plugin run operation.
+type RunPluginResult struct {
+	NodesUpserted     int `json:"nodesUpserted"`
+	RelationsUpserted int `json:"relationsUpserted"`
+}
+
 // OperationResource is the JSON representation of an AIP-151 operation.
 type OperationResource struct {
-	Name              string               `json:"name"`
-	Plugin            string               `json:"plugin"`
-	State             string               `json:"state"`
-	StartTime         time.Time            `json:"startTime"`
-	EndTime           *time.Time           `json:"endTime,omitempty"`
-	Error             *StatusErrorResource `json:"error,omitempty"`
-	NodesUpserted     int                  `json:"nodesUpserted"`
-	RelationsUpserted int                  `json:"relationsUpserted"`
-	CreatedAt         time.Time            `json:"createdAt"`
+	Name     string                    `json:"name"`
+	Done     bool                      `json:"done"`
+	Metadata OperationMetadataResource `json:"metadata"`
+	Response *RunPluginResult          `json:"response,omitempty"`
+	Error    *StatusErrorResource      `json:"error,omitempty"`
 }
 
 type ListOperationsResponse struct {
@@ -47,24 +59,31 @@ var operationListOptionsSpec = listOptionsSpec{
 }
 
 func operationFromCatalogOperation(op operations.Operation) OperationResource {
-	var statusErr *StatusErrorResource
-	if op.Error != nil {
-		statusErr = &StatusErrorResource{
-			Message: op.Error.Message,
-		}
+	done := op.State == operations.StateSucceeded || op.State == operations.StateFailed
+
+	resource := OperationResource{
+		Name: op.Name,
+		Done: done,
+		Metadata: OperationMetadataResource{
+			Plugin:    op.Plugin,
+			State:     string(op.State),
+			StartTime: op.StartTime,
+			EndTime:   op.EndTime,
+			CreatedAt: op.CreatedAt,
+		},
 	}
 
-	return OperationResource{
-		Name:              op.Name,
-		Plugin:            op.Plugin,
-		State:             string(op.State),
-		StartTime:         op.StartTime,
-		EndTime:           op.EndTime,
-		Error:             statusErr,
-		NodesUpserted:     op.NodesUpserted,
-		RelationsUpserted: op.RelationsUpserted,
-		CreatedAt:         op.CreatedAt,
+	switch {
+	case done && op.State == operations.StateSucceeded:
+		resource.Response = &RunPluginResult{
+			NodesUpserted:     op.NodesUpserted,
+			RelationsUpserted: op.RelationsUpserted,
+		}
+	case done && op.Error != nil:
+		resource.Error = &StatusErrorResource{Message: op.Error.Message}
 	}
+
+	return resource
 }
 
 func toOperationResources(ops []operations.Operation) []OperationResource {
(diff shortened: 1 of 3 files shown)


catalog/internal/httpapi/operations.go
```diff
@@ -12,23 +12,34 @@ import (
...
+}
+
 // OperationResource is the JSON representation of an AIP-151 operation.
 type OperationResource struct {
-	Name              string               `json:"name"`
-	Plugin            string               `json:"plugin"`
-	State             string               `json:"state"`
-	StartTime         time.Time            `json:"startTime"`
-	EndTime           *time.Time           `json:"endTime,omitempty"`
-	Error             *StatusErrorResource `json:"error,omitempty"`
```
akavel-reply: Please put them in the same order as in https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L121 for easier comparison/reference:


```suggestion
	Name     string                    `json:"name"`
	Metadata OperationMetadataResource `json:"metadata"`
	Done     bool                      `json:"done"`
	Error    *StatusErrorResource      `json:"error,omitempty"`
	Response *RunPluginResult          `json:"response,omitempty"`
```

catalog/internal/httpapi/operations.go
```diff
@@ -12,23 +12,34 @@ import (
...
+	CreatedAt time.Time  `json:"createdAt"`
+}
+
+// RunPluginResult is the successful result of a plugin run operation.
+type RunPluginResult struct {
+	NodesUpserted     int `json:"nodesUpserted"`
+	RelationsUpserted int `json:"relationsUpserted"`
+}
+
 // OperationResource is the JSON representation of an AIP-151 operation.
```
akavel-reply:
```suggestion
// OperationResource is the JSON representation of an AIP-151 operation.
// See https://github.com/googleapis/googleapis/blob/0c516dc746bccd2e0f29a44e4b4e72a216bfc82a/google/longrunning/operations.proto#L121
// for documentation of fields.
```

catalog/internal/httpapi/operations.go
```diff
@@ -12,23 +12,34 @@ import (
...
-	StartTime         time.Time            `json:"startTime"`
-	EndTime           *time.Time           `json:"endTime,omitempty"`
-	Error             *StatusErrorResource `json:"error,omitempty"`
-	NodesUpserted     int                  `json:"nodesUpserted"`
-	RelationsUpserted int                  `json:"relationsUpserted"`
-	CreatedAt         time.Time            `json:"createdAt"`
+	Name     string                    `json:"name"`
+	Done     bool                      `json:"done"`
+	Metadata OperationMetadataResource `json:"metadata"`
+	Response *RunPluginResult          `json:"response,omitempty"`
```
akavel-reply: Type name does not fulfill naming requirement described in https://github.com/googleapis/googleapis/blob/0c516dc746bccd2e0f29a44e4b4e72a216bfc82a/google/longrunning/operations.proto#L121 and also is **not even consistent with our own naming of httpapi types** (based on AIPs).

catalog/internal/httpapi/operations.go
```diff
@@ -47,24 +58,31 @@ var operationListOptionsSpec = listOptionsSpec{
...
-			Message: op.Error.Message,
-		}
+	done := op.State == operations.StateSucceeded || op.State == operations.StateFailed
+
+	resource := OperationResource{
+		Name: op.Name,
+		Done: done,
+		Metadata: OperationMetadataResource{
+			Plugin:    op.Plugin,
+			State:     string(op.State),
```
akavel-reply: do we need this field anymore? I would expect we don't

catalog/internal/httpapi/operations.go
```diff
@@ -47,24 +58,31 @@ var operationListOptionsSpec = listOptionsSpec{
...
-		NodesUpserted:     op.NodesUpserted,
-		RelationsUpserted: op.RelationsUpserted,
-		CreatedAt:         op.CreatedAt,
+	switch {
+	case done && op.State == operations.StateSucceeded:
+		resource.Response = &RunPluginResult{
+			NodesUpserted:     op.NodesUpserted,
+			RelationsUpserted: op.RelationsUpserted,
+		}
+	case done && op.Error != nil:
```
akavel-reply: what if `op.State == StateFailed` but `op.Error == nil`?

catalog/internal/httpapi/operations.go
```diff
@@ -12,23 +12,34 @@ import (
...
-	State             string               `json:"state"`
-	StartTime         time.Time            `json:"startTime"`
-	EndTime           *time.Time           `json:"endTime,omitempty"`
-	Error             *StatusErrorResource `json:"error,omitempty"`
-	NodesUpserted     int                  `json:"nodesUpserted"`
-	RelationsUpserted int                  `json:"relationsUpserted"`
-	CreatedAt         time.Time            `json:"createdAt"`
+	Name     string                    `json:"name"`
+	Done     bool                      `json:"done"`
+	Metadata OperationMetadataResource `json:"metadata"`
```
akavel-reply: I'm very surprised and confused why the following seems to be **still passing successfully** after the above change - I would expect it should be failing and you should have to fix it in this PR, but I see CI tests passing, this is very suspicious to me. Could you please dig down and find out why CI is passing?

https://github.com/naira-project/naira/blob/af5692a49cced05e9ecaf7948e281296375c9d8f/plugins/cmd/depl_calls_svc/integration_test.go#L171

akavel-reply (changes requested):

catalog/internal/httpapi/operations.go
libera13: Good catch. Renamed to RunPluginResponse

catalog/internal/httpapi/operations.go
libera13: Removed from OperationMetadataResource

catalog/internal/httpapi/operations.go
libera13: Good catch. I changed it to populate error message in that case

catalog/internal/httpapi/operations.go
libera13: As you saw, the CI was updated and the test failed on it. I updated the test afterwards

catalog/internal/httpapi/operations.go
```diff
@@ -12,34 +12,35 @@ import (
...
-	State     string     `json:"state"`
 	StartTime time.Time  `json:"startTime"`
 	EndTime   *time.Time `json:"endTime,omitempty"`
 	CreatedAt time.Time  `json:"createdAt"`
 }
 
-// RunPluginResult is the successful result of a plugin run operation.
-type RunPluginResult struct {
+// StatusErrorResource is an error representation carried by failed operations.
+type StatusErrorResource struct {
```
akavel-reply: Is there some reason why we call this "StatusErrorResource" rather than just "ErrorResource"?

akavel-reply (approved):

kocakkaan (approved):
Minimal changes in UI and backend looks nice. Looks good to me 👍

